Issue #5, opened 2026-04-16
FIRMWARE Planning and Discussion


morrownr:
I am opening this issue as a place to plan and work things related to the firmware directory. I am currently compiling a text document about firmware as it relates to this repo so that we have good ongoing documentation so that there is a good understanding in the Linux community. The idea is that the produced document is stored in the firmware directory.

As of this date, the firmware directory contains the most up to date firmware files for the most commonly used devices within the scope of this repo. It is intended for this repo to meet the needs of desktop, laptop, MiniPC and SBC users of USB, PCI and SDIO devices, however, the primary mission of this site is to support USB WiFi on Linux so that will take priority.

What is the desired result of me producing this document?

So that we can make initial decisions about what devices will have  firmware files in the firmware directory. These will be the files that will be maintained and set to upload into systems as `install-driver.sh` is run. It is possible that the code to update firmware files will be moved to a separate script called something like `update-firmware.sh` as that could be handy for users simply wanting to update their firmware files but a decision on that can wait to see what the community wants.

Info: Mediatek supports many devices on Linux and some of them are Linux only. Some can be compiled from this repo but are of little use to the mission of this site. Think about the devices such as SoCs for Wireless Routers or TVs or Smartphones. Do we need the firmware files for those use cases? I do not think so and I don't think the drivers need to be compiled... however, keeping the repo current with all patches that go into mt76 is, I think, a good idea as we can simply use commenting one line to not compile what we do not think that we need. That allows flexibility to change rapidly in case we find that something that we have inactivated, should be activated.

Your thoughts on this matter are very welcome. I will publish the document as soon as able.

Cheers,

@morrownr

Lucid-Duck:
@morrownr

Spent last evening and this morning building a citation-backed compendium of MediaTek consumer WiFi on Linux -- 78 SKU rows across USB, PCIe/M.2, and SDIO, anchored in kernel source, linux-firmware, linux-hardware.org probes, Lenovo/Dell/HP/Framework spec sheets, FCC grantee IDs, CVE databases, and retail channels. Full dataset at https://github.com/Lucid-Duck/mt76-market-footprint.

Firmware to ship in `morrownr/mt76`, strict priority order:

[...]

PR #8, opened 2026-04-18
info: add MARKET-FOOTPRINT.md

Lucid-Duck:
Picks up the thread from issue #5. Adding `info/MARKET-FOOTPRINT.md` with the full market-footprint research -- 78 SKU rows across USB, PCIe/M.2, and SDIO, anchored in kernel source, linux-firmware, linux-hardware.org probes, Lenovo/Dell/HP/Framework spec sheets, FCC grantee IDs, CVE databases, and retail channels.

This follows the directory shape from your comment: document in `info/`, firmware/README.md links to it. Content is copied from my external working repo with self-references stripped so this file is the canonical source going forward. I've added a snapshot notice to my external repo pointing back here as the evolving version, so there's a single place to update.

@morrownr ready for review whenever you're at a stopping point.

### DIFF
--- a/info/MARKET-FOOTPRINT.md
+++ b/info/MARKET-FOOTPRINT.md
@@ -1,0 +1,201 @@
+# MediaTek Linux WiFi Market Footprint
+
+A data-driven survey of which **MediaTek WiFi adapters and modules** are most commonly used by Linux users, across USB, PCIe/M.2, and SDIO form factors. Intended as input to the firmware-shipping decisions for [`morrownr/mt76`](https://github.com/morrownr/mt76).
+
+Every claim cites a primary source. Every figure is time-stamped. Kernel anchoring to torvalds/linux HEAD `d730905bc3c0`; linux-firmware to gitlab HEAD `3fc7117bb925`; morrownr/mt76 to `131771025d08` (all 2026-04-17).
+
+## The question
+
+[`morrownr/mt76` issue #5](https://github.com/morrownr/mt76/issues/5) asks, in essence: *what are the most commonly used devices within the scope of this repo?*
+
+## Scope
+
+- **MediaTek-chipset adapters and modules** for **desktop, laptop, mini-PC, SBC, gaming handheld, and Chromebook** Linux users.
+- Form factors covered: USB, PCIe / M.2, SDIO.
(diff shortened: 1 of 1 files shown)


morrownr:
@Lucid-Duck 

Merged. Excellent document.

I did see a couple of things that did not line up with my memory but they were minor details and I need to research them and let you know as able. Example:

It was noted that mt7921 support was available as a backport on Debian 12. I do not think that is correct as the mt7921 driver/firmware available as of kernel 5.18 and Debian 12 used kernel 6.1. My memory may not be correct.

Lucid-Duck:
Good catch. Verified against torvalds master via gh api: mt7921e (PCIe)
landed in 5.12 via commit 5c14a5f944, mt7921s (SDIO) in 5.16 via
48fab5bbef, mt7921u (USB) in 5.18 via 0d2afe09fa. Your 5.18 figure
matches the USB variant; the PCIe and SDIO variants predate Debian 12's
6.1 kernel by years, so Debian 12 has all three natively.

Same check flagged a second table error while I was there: MT7922 shows
5.17 in the mainline-kernel column but the actual intro commit
688088728b first lands in 5.16. The Debian-12 cell for MT7922 reads
"yes" already so the user-facing answer isn't wrong, but the kernel
number should still be corrected.

I'll push a small PR fixing both cells. If more turn up as you dig, let
me know and I can fold them into the same PR.

morrownr:
I'll take the time to do a more detailed review later today.

While you are making patches... since we are in agreement on what firmware to maintain and what files to compile, the following can be commented out in Makefile:

obj-m += mt7615/
obj-m += mt7915/
obj-m += mt7996/

I was going over install-driver.sh and ininstall-driver.sh this morning. Glad you could recycle some of the existing code. Some of my code probably hurt your head, Ha ha...

Lucid-Duck:
mt7915 and mt7996 commented out in #11.

Held mt7615 back: that subdir also builds mt7663 (mt7663_usb_sdio_common,
mt7663u, mt7663s), which the firmware README keeps as "firmware included,
driver compiled" -- you just added the _rebb variants this morning.
Commenting out `obj-m += mt7615/` would drop mt7663 too.

Two reasonable paths: leave mt7615/ active (simplest -- mt7615_common
and mt7615e come along as fellow travelers), or split mt7663 into its
own subdir so mt7615/mt7615e can go dark on their own. Your call.

PR #9, opened 2026-04-19
firmware: fix README typos and add missing MT7925 Status line

Lucid-Duck:
Four small corrections to `firmware/README.md` following up on #5 (closed) where you mentioned you'd circle back to check editing.

1. **`MARKET-FOOTPRINT.md0)` link** -- trailing `0` before the closing paren. URL 404s. Changed to `MARKET-FOOTPRINT.md)`.
2. **MT7921 header** -- read `2.4/5GHz/6Hz`. Changed to `2.4/5/6GHz` to match the band convention used everywhere else in the file.
3. **MT7925 firmware filename** -- read `WIFI_RAM_CODE_MT7925_1_1.binBB`. Changed to `WIFI_RAM_CODE_MT7925_1_1.bin` (verified against `linux-firmware.git`).
4. **MT7925 `Status:` line** -- MT7925 was the only chip block without one. Added `Status: firmware included, driver compiled` to match every other block.

No other changes. Let me know if you'd like any further nits folded in before merge.

### DIFF
--- a/firmware/README.md
+++ b/firmware/README.md
@@ -13,7 +13,7 @@ of this document and where updated files can be obtained:
 
 [Mediatek Official Linux firmware repository](https://git.kernel.org/pub/scm/linux/kernel/git/firmware/linux-firmware.git/tree/mediatek)
 
-Information regarding the mt76 [MARKET-FOOTPRINT](https://github.com/morrownr/mt76/blob/main/info/MARKET-FOOTPRINT.md0)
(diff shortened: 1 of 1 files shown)


morrownr:
Thanks for this. I do need my stuff edited as I simply do not see as well as I used to so I miss more than I should.

Lucid-Duck:
No trouble at all. I'll keep an editorial eye on anything that lands going forward so the second pass just happens in the background.

PR #11, opened 2026-04-19
Makefile: stop compiling mt7915 and mt7996 by default

Lucid-Duck:
Per #8: commenting out `obj-m += mt7915/` and `obj-m += mt7996/`.
Both chips are "no firmware, not compiled" in the README, so this
matches the existing mt7603/ treatment. Source trees stay so anyone
with matching hardware can uncomment locally.

mt7615/ left active -- flagging separately on #8 because dropping
it would also remove mt7663.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -81,16 +81,18 @@ CFLAGS_mt76x02_trace.o := -I$(src)
 CFLAGS_mt792x_trace.o := -I$(src)
 
 # --- Chipset families (subdirectories) ---
-# MT7603 (WiFi 4) is not built by default. The chip is old, no users have
-# reported running it on this out-of-tree driver, and no firmware is shipped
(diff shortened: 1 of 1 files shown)


morrownr:
@Lucid-Duck 

> mt7615/ left active -- flagging separately on #8 because dropping it would also remove mt7663.

Okay, I was unaware of that as neither the mt7663 or mt7615 have been on my radar. My attention has been mostly on USB and especially on products that are readily available for folks around the world but no products that I am aware of have met that criteria.

To keep things consistent, I will push the firmware files for mt7615 and make the needed changes in firmware/README.md.

Lucid-Duck:
Cleanest resolution -- thanks. Ping me if you want an editorial pass on the firmware/README.md update when it lands.

Issue #4, opened 2026-04-15
Compilation fails on kernels < 6.17


morrownr:
This is a known issue. I am posting this issue more or less as a to-do item and a place that we can gather information. The current idea is to backport down to and including kernel 6.12. Backporting beyond kernel 6.12 may be a challenge.

Here is the output from `install-driver.sh`:

```
$ sudo sh install-driver.sh
Please copy and post all displayed lines when reporting an issue!
Press Enter to continue...

  ================================================================
     mt76 WiFi Driver Installer
  ================================================================
  install-driver.sh v20260408
  Source:  25cfd622
  Distro:  Debian GNU/Linux 13 (trixie)
  Kernel:  6.12.75+rpt-rpi-v8 (aarch64)
  Params:  coherent_pool=1M, 8250.nr_uarts=1, snd_bcm2835.enable_headphones=0, cgroup_disable=memory, numa_policy=interleave, nvme.max_host_mem_size_mb=0, snd_bcm2835.enable_headphones=1, snd_bcm2835.enable_hdmi=1, snd_bcm2835.enable_hdmi=0, numa=fake=2, system_heap.max_order=0, iommu_dma_numa_policy=interleave, smsc95xx.macaddr=DC:A6:32:93:0D:0F, vc_mem.mem_base=0x3ec00000, vc_mem.mem_size=0x40000000, console=ttyAMA0,115200, console=tty1, rootfstype=ext4, fsck.repair=yes, rootwait, quiet, splash, plymouth.ignore-serial-consoles, cfg80211.ieee80211_regdom=US
  CPUs:    4/4 (in-use/total)
  Memory:  3887804 kB
  gcc:     gcc (Debian 14.2.0-19) 14.2.0
  SecBoot: unknown (mokutil not installed)
  Reg:     country US: DFS-FCC

  Detected mt76 hardware:
    mt7925u: ID 0846:9072 MediaTek Inc. Wireless_Device
  ----------------------------------------------------------------

[1/7] Checking prerequisites
  All prerequisites found.
  Kernel headers present.

[2/7] Removing previous installation (if any)
  Clean.

[3/7] Configuring
  Installing mt76_git.conf to /etc/modprobe.d
  Done.

[4/7] Installing firmware
  Installed 16 firmware file(s) to /lib/firmware/mediatek

[5/7] Building modules
  (non-dkms build with 4 cores)
make -j3 -C /lib/modules/6.12.75+rpt-rpi-v8/build M=$PWD modules
[...]

PR #12, opened 2026-04-19
wifi: mt76: add kernel 6.13-6.17 compat for mac80211 ieee80211_ops API changes

Lucid-Duck:
Scope B of #4. Completes the 6.12-to-6.16 kernel backport by covering
every `ieee80211_ops` API break between v6.12 and v6.17 that Scope A
(`13177102`) did not.

- **6.13** -- `sta_rc_update` renamed to `link_sta_rc_update` (third
  arg type flips to `ieee80211_link_sta *`). `struct wireless_dev`
  also gained `radio_mask`; pre-6.13 is single-radio via a small
  compat macro.
- **6.14** -- `get_txpower` gained `unsigned int link_id`.
- **6.15** -- `IEEE80211_HW_NO_VIRTUAL_MONITOR` guarded at
  `mt7996/init.c:539`. Closes the commitment to @ohquait on #6.
- **6.17** -- `int radio_idx` added to six callbacks (config,
  set_frag_threshold, set_rts_threshold, set_coverage_class,
  set_antenna, get_antenna). 29 impl sites + 5 header prototypes
  guarded.

Pattern matches Scope A: inline `#if LINUX_VERSION_CODE` with
`compat: ...` tag comments so the full surface is greppable and can
be removed cleanly when the minimum supported kernel rises past
these cutoffs.

Tested: Pi5 6.12.47 aarch64, debian 6.12 / ubuntu 6.17 / kali 6.18 /
alpine 6.18 (musl+gcc15) / fedora 6.19 x86_64. All compile and
load/unload clean, taints OOT-only. Runtime smoke (mon0 + live frame
capture) on Pi5 (native USB) and a debian 6.12 VM (USB passthrough)
with an Alfa AWUS036AXML -- real over-the-air beacons captured,
3 rounds of mon0 cycle stress with no oops/WARN.

### DIFF
--- a/mac80211.c
+++ b/mac80211.c
@@ -1778,8 +1778,14 @@ s8 mt76_get_power_bound(struct mt76_phy *phy, s8 txpower)
 }
 EXPORT_SYMBOL_GPL(mt76_get_power_bound);
 
+/* compat: get_txpower gained link_id in kernel 6.14 */
+#if LINUX_VERSION_CODE >= KERNEL_VERSION(6, 14, 0)
 int mt76_get_txpower(struct ieee80211_hw *hw, struct ieee80211_vif *vif,
 		     unsigned int link_id, int *dbm)
+#else
+int mt76_get_txpower(struct ieee80211_hw *hw, struct ieee80211_vif *vif,
+		     int *dbm)
+#endif
 {
 	struct mt76_phy *phy = mt76_vif_phy(hw, vif);
 	int n_chains, delta;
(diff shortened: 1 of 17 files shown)


morrownr:
Wow. You are fast. Doing this in the 2 Realtek out-of-kernel drivers took me a little time but that was only 2 drivers. That was a lot of work. Your work is appreciated. I think this repo is going to prove to be very helpful to the Linux wireless community.

Lucid-Duck:
Thanks. The compat pattern turned out to be mechanical once the full API diff was in hand -- the real time was getting proof across 6 kernels and 2 arches before sending. Good to have it in the tree.

PR #13, opened 2026-04-19
info: drop RHEL columns from MARKET-FOOTPRINT distro matrix

Lucid-Duck:
Per your comment on #10: Red Hat's selective cherry-pick across
kernels makes "what's supported on RHEL N" unreliable to state in a
general matrix. Removes RHEL 10.1 and RHEL 9 columns, updates the
bottom-line paragraph to note why (so a future reader who wonders
why RHEL is absent has the answer in place).

Other Red Hat mentions -- the bugzilla reference and the MT7925
linux-firmware Red Hat BZ under the sentiment section -- are
preserved since those are bug-tracker signal, not support claims.

### DIFF
--- a/info/MARKET-FOOTPRINT.md
+++ b/info/MARKET-FOOTPRINT.md
@@ -80,17 +80,17 @@ Confirmed non-MediaTek: Lenovo ThinkStation P3 Tower/Tiny/Ultra + P5 + P7 + P8 (
 
 ### Distro LTS kernel landing matrix (2026-04-17)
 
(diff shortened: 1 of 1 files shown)


morrownr:
Probably a good call.

BTW, you forgot to edit the main README to note the change in supported kernels... 6.12 to 7.x

Lucid-Duck:
Derp, my bad. That's what I get for going too fast! PR #14 to follow with the main README kernel range fix, plus the compat-patches/ description that was also stale.

PR #14, opened 2026-04-19
README: update supported kernel range from 6.17+ to 6.12+ post-Scope-B

Lucid-Duck:
Per #13 comment: the main README still claimed "kernels 6.17 through
7.x" in three places after Scope A (#4) and Scope B (#12) merged.
Real supported floor is 6.12.

- Top summary line: 6.17 -> 6.12.
- Kernel Compatibility section: same.
- `compat-patches/` row in file-map table was also stale -- said
  "for kernels 6.17 through 6.19+", but that directory's one patch
  is for 6.6+, and the Scope A/B inline fences now carry the real
  compat load. Replaced with a generic "older kernel API changes"
  description.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 # mt76 Out-of-Tree WiFi Driver Build Infrastructure
 
 Build, install, and manage MediaTek mt76 WiFi drivers as out-of-tree kernel
-modules. Supports DKMS, non-DKMS, Secure Boot, and kernels 6.17 through 7.x.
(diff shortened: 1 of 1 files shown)


morrownr:
Thanks. Time to let the smoke clear.

I plan to continue with testing 6.12+ tonight.

I guess the next thing on the agenda would be to consider which is the patches for kernel 7.2 that are not accepted yet might need some testing... maybe. And we need to plan to plug patches from openwrt/mt76 so as to apply them here. I have only taught myself what I needed to know about git to handle the REALTEK out-of-kernel repo so I could use an education if you have time...

Issue #21, opened 2026-04-28
Failure to build on 7.1-rc1


satmandu:
```
sudo ./install-driver.sh
Please copy and post all displayed lines when reporting an issue!
Press Enter to continue...

  ================================================================
     mt76 WiFi Driver Installer
  ================================================================
  install-driver.sh v20260408
  Source:  90ab9ae0
  Distro:  Ubuntu 26.04 LTS
  Kernel:  7.1.0-rc1 (x86_64)
  Params:   ro, mitigations=off, i915.mitigations=off, acpi_mask_gpe=0x17, rcuupdate.rcu_cpu_stall_suppress=1, zswap.enabled=1, zswap.zpool=zsmalloc, log_buf_len=1M, nouveau.config=NvPmEnableGating=3, zfs.zil.replay_disable=1, zfs.zfs_recover=1, cfi=off, nouveau.atomic=1
  CPUs:    8/8 (in-use/total)
  Memory:  16279380 kB
  gcc:     gcc (Ubuntu 15.2.0-16ubuntu1) 15.2.0
  dkms:    dkms-3.2.2
  SecBoot: not supported
  Reg:     country US: DFS-FCC

  Detected mt76 hardware:
    mt7925u: ID 0846:9072 MediaTek Inc. Wireless_Device
  ----------------------------------------------------------------

[1/7] Checking prerequisites
  All prerequisites found.
  Kernel headers present.

[2/7] Removing previous installation (if any)
  Clean.

[3/7] Configuring
  Installing mt76_git.conf to /etc/modprobe.d
  Done.

[4/7] Installing firmware
  Installed 23 firmware file(s) to /lib/firmware/mediatek

[5/7] Building modules
  (dkms build)
  Copying source to /usr/src/mt76-1.0
Creating symlink /var/lib/dkms/mt76/1.0/source -> /usr/src/mt76-1.0
  Added to dkms.
Sign command: /usr/bin/kmodsign
Signing key: /var/lib/shim-signed/mok/MOK.priv
Public certificate (MOK): /var/lib/shim-signed/mok/MOK.der

Building module(s).........(bad exit status: 2)
Failed command:
KVER=7.1.0-rc1 'make' modules
ERROR (dkms apport): binary package for mt76: 1.0 not found

Error! Bad return status for module build on kernel: 7.1.0-rc1 (x86_64)
Consult /var/lib/dkms/mt76/1.0/build/make.log for more information.
Command exited with non-zero status 10
[...]

PR #24, opened 2026-04-29
install/uninstall-driver.sh: verbose firmware install, initramfs rebuild step

Lucid-Duck:
Two ergonomic improvements to the install / uninstall flow:

1. **Step 4 (Installing firmware)** prints each file and destination instead of just the count summary. Per the MT7927 mail thread; helps when "firmware did not install correctly" and you need to know which file went where. Same Step 4 also iterates any chip-specific subdirectory under `firmware/` (mt7925, mt7927, future) generically instead of hardcoding mt7925.

2. **New install Step 8 / uninstall Step 5** regenerates the initramfs via the available tool: `dracut` (Fedora / RHEL / openSUSE / Arch+dracut), `update-initramfs` (Debian / Ubuntu), or `mkinitcpio` (standard Arch). Per @satmandu's note in #21: without this, the initramfs can keep loading the in-tree mt76 modules from a stale image despite the modprobe.d blacklist, so the install appears not to stick after reboot.

Tested end-to-end in the fedora-mt76 VM (dracut path, full install + uninstall cycle). ShellCheck clean.

### DIFF
--- a/install-driver.sh
+++ b/install-driver.sh
@@ -65,7 +65,7 @@ else
 fi
 
 # helper: step counter
-TOTAL_STEPS=7
+TOTAL_STEPS=8
 step() {
 	printf '\n%s%s[%s/%s]%s %s%s%s\n' "${BOLD}" "${CYAN}" "$1" "${TOTAL_STEPS}" "${NC}" "${BOLD}" "$2" "${NC}"
 }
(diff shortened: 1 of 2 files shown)


satmandu:
FYI, Ubuntu can now also use dracut. (I'm using dracut on Ubuntu 26.04...)

Lucid-Duck:
Thanks @satmandu -- the detection order is `dracut → update-initramfs → mkinitcpio`, so dracut takes precedence when present, which covers your Ubuntu 26.04 rig. Verified post-merge across all four paths (dracut, update-initramfs, mkinitcpio, and the no-known-tool warning fallback for distros like Alpine).

PR #30, opened 2026-05-08
install: split parameters edit out into edit-parameters.sh

Lucid-Duck:
Per your suggestion: the mid-install "edit driver options now?" prompt is residue from the older Realtek driver tooling and is not pulling its weight on mt76, where the defaults work for the vast majority of MT7921/7922/7925 users.

This PR:

- Adds `edit-parameters.sh` mirroring `rtl8852bu-20250826/edit-options.sh` shape, renamed because `/etc/modprobe.d/mt76_git.conf` holds module parameters not options.
- Adds `default-editor.txt` (`nano`) for the same fallback behaviour as the rtl repos.
- Removes the inline editor-discovery block and the "edit driver options now?" prompt from `install-driver.sh`. Users who want to tweak parameters now run `sudo ./edit-parameters.sh` after the install completes.
- Adds an entry in the install-driver.sh success banner pointing at `edit-parameters.sh`.
- Adds the explanatory sentences above the reboot prompt that we discussed, so users understand why a reboot is the simplest path to apply parameter changes.
- Documents the new script in `README.md` (Quick Start + Files table).

Both scripts are POSIX sh, shellcheck clean, sh -n clean, and the sudo guard on edit-parameters.sh exits 1 correctly without root.

Ready for your review whenever you have the eye for it.


morrownr (commented):
Looks good. Your attention to detail is very good. Thanks.

PR #33, opened 2026-05-12
install: drop pre-banner 'copy and post' prompt

Lucid-Duck:
Per #32: drops the pre-banner "Please copy and post all displayed lines when reporting an issue! Press Enter to continue..." prompt block from `install-driver.sh`.

`check-driver.sh` now produces a structured diagnostic that the prompt was steering users toward, so the line is dead UX. The `Press Enter` pause also blocks unattended installs (cron, cloud-init, CI runners).

The `NO_PROMPT` detection at the top of the script is kept intact because the post-install prompts at the bottom still rely on it.

Six-line removal, no behaviour change other than the banner now appearing immediately on launch.

### DIFF
--- a/install-driver.sh
+++ b/install-driver.sh
@@ -124,12 +124,6 @@ fi
 # ===========================================================================
 # Banner
 # ===========================================================================
-if [ $NO_PROMPT -ne 1 ]; then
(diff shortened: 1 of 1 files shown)


morrownr:
Awesome.

PR #42, opened 2026-06-10
mt792x: restore < 6.14 compat for mt792x_get_txpower

Lucid-Duck:
`mt792x_get_txpower` gained the `unsigned int link_id` parameter in `8adc84f5` ("wifi: mt76: mt792x: report txpower for the requested vif link"), an upstream patch with the unconditional signature. The `< 6.14` compat gate that already protects `mt76_get_txpower` in `mt76.h` wasn't extended to the `mt792x` wrapper, so the tree builds on >= 6.14 but fails below 6.14:

    mt792x_core.c:351: error: too many arguments to function 'mt76_get_txpower'
    mt792x_core.c:351: error: passing argument 3 ... makes pointer from integer [-Wint-conversion]

This gates `mt792x_get_txpower` (the definition, the `mt76_get_txpower` call, and the prototype) at the same `LINUX_VERSION_CODE >= KERNEL_VERSION(6, 14, 0)` boundary `mt76.h` uses. On < 6.14 it drops `link_id` and uses link 0 (deflink), matching the pre-MLO callback signature.

Bisected on a Pi 5 (6.12.47, gcc 14): the parent commit (`cc42fbb`) builds clean, `8adc84f5` breaks it, this PR builds clean again (26 modules). 7.0.11 (gcc 16) unchanged.

### DIFF
--- a/mt792x.h
+++ b/mt792x.h
@@ -418,8 +418,13 @@ void mt792x_roc_timer(struct timer_list *timer);
 void mt792x_csa_timer(struct timer_list *timer);
 void mt792x_flush(struct ieee80211_hw *hw, struct ieee80211_vif *vif,
 		  u32 queues, bool drop);
+#if LINUX_VERSION_CODE >= KERNEL_VERSION(6, 14, 0)
 int mt792x_get_txpower(struct ieee80211_hw *hw, struct ieee80211_vif *vif,
 		       unsigned int link_id, int *dbm);
+#else
+int mt792x_get_txpower(struct ieee80211_hw *hw, struct ieee80211_vif *vif,
+		       int *dbm);
+#endif
 int mt792x_assign_vif_chanctx(struct ieee80211_hw *hw,
 			      struct ieee80211_vif *vif,
 			      struct ieee80211_bss_conf *link_conf,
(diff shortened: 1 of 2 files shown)


morrownr:
I had noted this problem while testing on 6.12 early this morning but I had to be out of the office for some time and you had it fixed by the time I returned. Thank you very much.

Side question: I need to learn some of the terminology you are using. I have been using the follow constructs for years:

```
#if LINUX_VERSION_CODE >= KERNEL_VERSION(6, 14, 0)

#else

#endif

```

Do you call this a wrapper? Help me out so that we are on the same sheet of music.

Lucid-Duck:
@morrownr glad that one's behind us, and good that you'd already caught it on 6.12.

On the wording: that `#if ... #else ... #endif` block is usually just called a version guard (or a version check). All it's doing is telling the compiler "build this on newer kernels, build that on older ones."

A wrapper is a different thing. That's when the kernel changes the shape of a function, say it grows an extra argument, and you write a little helper around it so the rest of the code doesn't have to care which kernel it's on. The guard picks which path to take; the wrapper is the small adapter that does the actual smoothing-over.

They often work together, but the mt792x fix was really just the guard: two copies of the one function chosen by kernel version, no separate wrapper needed.

PR #46, opened 2026-06-20
[codex] Add MT7927 monitor and MLO support

Ashcal9669:
Adds MT7927 PCIe support as a standalone mt7927 module family, including monitor-mode 
    band handling and shared Connac queue handling needed for monitor injection traffic.

Changes:
  - add mt7927 PCIe driver sources and build integration
  - keep MT7927 PCI IDs out of the MT7925 PCI driver so MT7927 binds to mt7927e_git
  - preserve non-AC TX queue mappings in the Connac PM queue path by widening pending TX tracking 
  to __MT_TXQ_MAX
  - set MT7927 sniffer band indexes from the active channel context
  - use real monitor interfaces for MT7927 instead of virtual monitor VIFs
  - document fresh install and monitor-mode setup in docs/mt7927-install.md
  - This also aligns the MT7927 monitor-mode path with Sean Wang's real-monitor-VIF direction: MT7927 uses a real monitor interface instead of relying on a virtual monitor VIF, and the sniffer band index is selected from the active channel context.
  
Validation:
  - make -j4 passed on 7.0.10+deb13-amd64
  - fresh reboot MLO validation completed association on one active 6 GHz / 320 MHz MLO link
  - monitor-mode validation used the real netdev converted to monitor mode; hcxdumptool captured
  traffic with zero kernel drops in the successful run

  Claim boundary:
  MLO is verified here as one active associated 6 GHz / 320 MHz MLO link, not full tri-band
  simultaneous MLO.


morrownr:
@Lucid-Duck 

I assigned you to provide a review but really, what I intend to communicate is that I will not be doing anything with this PR unless you tell me to do something.

Lucid-Duck:
Thanks for putting the work in here, @Ashcal9669.

I'm going to pass on this one, but let me explain why so it's not just a no. The MT7927 is basically the MT7925's twin. Same chip family, so the driver already handles it inside mt7925 today (the 7927 IDs already live in mt7925's PCI table). That's how everyone does it: openwrt, the mainline kernel, and the MT7927 patches going up on the linux-wireless list right now all add it to mt7925 instead of giving it a driver of its own. This PR goes the other way. It pulls 7927 back out of mt7925 and stands up a separate mt7927 module that is mostly a copy of mt7925 with the names changed. That moves us away from upstream and gets painful to maintain fast, so I can't take it as it stands.

There is real work in here though, and I went through it properly. Underneath the duplicated module it is a few hundred lines of genuine changes, and they all apply straight to mt7925: the monitor-mode channel and band handling, the MLO link selection, the EML capability detection in the MCU path, the sniffer band index, and keeping the non-AC TX queue mapping in the power-save path instead of forcing everything to BE. You clearly put it on real hardware too, with an MLO link up and a clean monitor capture, which is exactly why it is worth landing properly. If you send those as separate patches against mt7925 I'll go through them one at a time. The monitor and MLO parts are right where my attention is.

One heads up: a couple of those touch the shared connac code, so they will need a check on mt7921 and mt7925 too, not just 7927. Routine, just flagging it.

Ashcal9669:
Understood. I will rework this as focused patches against the existing `mt7925` path instead of keeping a standalone `mt7927` module.

I’ll split the work roughly as:

  1. MT7927 sniffer / monitor band-index handling in `mt7925`
  2. monitor-mode channel handling aligned with the real-monitor-VIF direction
  3. MLO / EML capability and link-selection fixes
  4. Connac PM queue mapping preservation, with checks on MT7921 and MT7925 impact

I’ll keep the validation boundary explicit: current MLO evidence is one active associated 6 GHz / 320 MHz MLO link, plus the clean monitor-mode hcxdumptool capture. I’ll also run build/regression checks before opening the replacement PRs.

Ashcal9669:
Clarifying the rework: the replacement patches will stay on the existing `mt7925` path and follow the same direction Sean Wang is taking upstream. The monitor work keeps the real monitor-interface handling, and the MLO / EML fixes stay as MT7927-gated behavior inside `mt7925`, not as a standalone `mt7927` driver.

PR #47, opened 2026-06-21
[codex] mt76: preserve non-AC TX queues during Connac PM wake

Ashcal9669:
Preserve non-AC TX queue mapping in the shared Connac PM wake path by widening the pending TX tracking to __MT_TXQ_MAX and keeping mt7925/mt7921 queue handling intact.

This is split out from the MT7927 rework and stays on the shared mt76/mt792x path so it can be reviewed against the existing mt7925 direction instead of a standalone mt7927 driver.

Validation:
  - make -j4

### DIFF
--- a/mt76_connac.h
+++ b/mt76_connac.h
@@ -95,7 +95,7 @@ struct mt76_connac_pm {
 	struct {
 		struct mt76_wcid *wcid;
 		struct sk_buff *skb;
-	} tx_q[IEEE80211_NUM_ACS];
+	} tx_q[__MT_TXQ_MAX];
 
 	struct work_struct wake_work;
 	wait_queue_head_t wait;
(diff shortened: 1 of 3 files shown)


Lucid-Duck:
Thanks for getting these up so quickly. I'll work through #47, #48 and #49 today.

Lucid-Duck:
I had a proper look at this one, and I don't think it does anything.

The queue value coming into mt792x_tx is always a normal mac80211 access category (0 to 3), so it never reaches the case the clamp is guarding, and the extra queue slots it adds never get used. Anything that did slip through gets put back to BE in mt76_tx anyway. I traced that path on real mt792x silicon while pushing traffic and cycling power save, and the queue function this touches wasn't called once, and the existing WARN never fired. Mainline and openwrt both keep the clamp this drops. It's shared connac code, so the MT7927 I've been testing runs the exact same path, nothing chip-specific changes here.

So I'd leave it, unless you've got a real capture showing a frame getting collapsed, in which case send it over and I'll look again.

morrownr:
Per conversation with @Lucid-Duck , #47 is on hold.

morrownr:
I've been busy. I'm closing this patch. Let me know if I missed something.

PR #48, opened 2026-06-21
[codex] mt76: set MT7927 monitor sniffer band

Ashcal9669:
Teach the MT7925 monitor path to select the MT7927 sniffer band from the active channel context so real monitor mode uses the right band index during capture and injection.

This stays aligned with the existing mt7925 direction and the real-monitor-interface approach discussed in review.

Validation:
  - make -j4

### DIFF
--- a/mt7925/main.c
+++ b/mt7925/main.c
@@ -771,12 +771,36 @@ mt7925_pm_interface_iter(void *priv, u8 *mac, struct ieee80211_vif *vif)
 }
 
 static void
+mt7925_monitor_update_chan(struct mt792x_vif *mvif,
+			   struct ieee80211_chanctx_conf *ctx)
+{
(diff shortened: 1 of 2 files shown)


Lucid-Duck:
Tested this on a real MT7927 (Filogic 380) today, and it works.

Without the patch, monitor mode only hears the band the radio started on, and 2.4 GHz captures nothing. With it, the sniffer follows the channel you set, 2.4 GHz starts working, and switching back and forth is clean. It's all behind the is_mt7927 check, so 7921 and 7925 aren't touched. I'd take it.

One thing I noticed that's separate from this patch: 5 and 6 GHz share the same band index, and in monitor mode the radio won't move between them unless you re-associate on the one you want (same both directions). The band index is already right, so it isn't this patch, but it's worth a look later if you want monitor to hop freely across all the bands.

PR #49, opened 2026-06-21
[codex] mt76: parse MT7927 MLO EML capabilities

Ashcal9669:
Keep the MT7925 MLO path aware of the MT7927 EML capability shape, with MT7927-gated capability parsing and negotiated EMLSR handling inside the existing mt7925 code path.

This follows the same upstream direction as the other mt7925-based MT7927 work rather than a standalone mt7927 module.

Validation:
  - make -j4
  - one active associated 6 GHz / 320 MHz MLO link on fresh reboot

### DIFF
--- a/mt7925/main.c
+++ b/mt7925/main.c
@@ -284,6 +284,11 @@ int mt7925_init_mlo_caps(struct mt792x_phy *phy)
 	ext_capab[0].eml_capabilities = phy->eml_cap;
 	ext_capab[0].mld_capa_and_ops =
 		u16_encode_bits(0, IEEE80211_MLD_CAP_OP_MAX_SIMUL_LINKS);
+	if (is_mt7927(phy->mt76->dev)) {
+		ext_capab[0].mld_capa_and_ops =
+			u16_encode_bits(1, IEEE80211_MLD_CAP_OP_MAX_SIMUL_LINKS) |
+			IEEE80211_MLD_CAP_OP_FREQ_SEP_TYPE_IND;
+	}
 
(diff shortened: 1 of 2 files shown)


Lucid-Duck:
Tested this on a real MT7927 today with a capability dump, and the EML parsing is reading the wrong bytes.

The first two parts look fine. The part I'd push back on is the parse_eml_cap change. On the 7927 the EML data is the same 12-byte layout that's documented, and the eml_cap value at offset 4 reads 0x0000, exactly like the 7925. The new code reads offset 2 instead, which sits in the reserved area, and on both chips those bytes happen to have the bottom bit set (0x0001 on the 7927, 0x0201 on the 7925). That switches EMLSR on even though the firmware reports it off. On the 7925 that's a real change to a shipping chip, so I'd put that read back to the documented offset 4.

That said, the chip_cap part is fixing something real. On the firmware I have here (build 20260414), the 7927 doesn't report a CHIP_CAP entry at all, so chip_cap stays 0, the MLO setup bails out early, and the 7927 ends up with no MLO. So turning MLO on for the 7927 is the right call. I'd just put it behind the is_mt7927 check and set MLO_EN there, and leave EMLSR alone for now, since the field reads 0. If the 7927 really does support EMLSR, I'd rather switch it on from a clear signal than from the reserved bytes.

Ashcal9669:
Updated PR #49 to restore documented 12-byte EML layout and gated the MLO fallback to MT7927-only hardware. I am now proceeding with porting the remaining monitor/injection logic into the mt7925 integration path as requested.

PR #51, opened 2026-07-01
[codex] mt7927: stabilize monitor lifecycle

Ashcal9669:
Stabilizes MT7927 monitor mode after repeated filter transitions and channel changes.\n\nWhat changed:\n- re-arm the MT7927 sniffer path after filter updates so monitor mode keeps producing frames across repeated capture runs\n- keep the MT7927 monitor chandef in sync with the active channel context\n- reject stale monitor frames whose reported frequency does not match the active MT7927 chandef\n- propagate the active MT7927 band index into the shared RX filter path\n- keep module install compression compatible with distros that store modules as *.ko.xz\n\nWhy:\n- this fixes the monitor lifecycle regression that showed up after the 7.0.12+deb13-amd64 kernel upgrade, where the card could enter monitor mode but stop producing useful RX after repeated open/close cycles\n\nValidation:\n- built successfully on 7.0.12+deb13-amd64

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -151,7 +151,7 @@ install:
 		zstd -fq --rm $(MODDIR)/*.ko 2>/dev/null || true; \
 	elif ls /lib/modules/$(KVER)/kernel/net/wireless/*.ko.xz >/dev/null 2>&1; then \
 		echo "Compressing modules with xz (matching distro scheme)..."; \
-		xz -f $(MODDIR)/*.ko 2>/dev/null || true; \
+		xz -f -C crc32 $(MODDIR)/*.ko 2>/dev/null || true; \
 	elif ls /lib/modules/$(KVER)/kernel/net/wireless/*.ko.gz >/dev/null 2>&1; then \
 		echo "Compressing modules with gzip (matching distro scheme)..."; \
 		gzip -f $(MODDIR)/*.ko 2>/dev/null || true; \
--- a/mt7925/mac.c
+++ b/mt7925/mac.c
@@ -406,6 +406,13 @@ mt7925_mac_fill_rx(struct mt792x_dev *dev, struct sk_buff *skb)
 
 	mt792x_get_status_freq_info(status, chfreq);
 
+	/* MT7927 reports frames from both DBDC RX chains in monitor mode. */
+	if (is_mt7927(&dev->mt76) &&
+	    (mphy->hw->conf.flags & IEEE80211_CONF_MONITOR) &&
+	    phy->mt76->chandef.chan &&
+	    status->freq != phy->mt76->chandef.chan->center_freq)
+		return -EINVAL;
+
 	switch (status->band) {
 	case NL80211_BAND_5GHZ:
 		sband = &mphy->sband_5g.sband;
--- a/mt7925/main.c
+++ b/mt7925/main.c
@@ -789,6 +789,9 @@ mt7925_monitor_update_chan(struct mt792x_vif *mvif,
 	if (!chan)
 		return;
 
+	if (ctx)
+		mvif->phy->mt76->chandef = ctx->def;
+
 	mconf->mt76.band_idx = mt7927_band_idx(chan->band);
 	mconf->mt76.basic_rates_idx = MT792x_BASIC_RATES_TBL;
 	if (chan->band != NL80211_BAND_2GHZ)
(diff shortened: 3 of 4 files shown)


Lucid-Duck:
@Ashcal9669 I built this on Fedora 7.0.12 and tested it on real MT7927 hardware (14c3:7927). It's clean and regression-free, monitor RX is healthy on 2.4/5/6 GHz across repeated open/close cycles and band switches.

But I can't reproduce the problem it targets. On the current base (with #48) I ran the same open/close cycling and a rapid filter-toggle stress and monitor RX never stopped, it kept delivering frames throughout. I also can't find the regression reported anywhere: no issue, and nothing in #18 or #20.

Before this goes in, can you post an exact reproduction on top of the merged #48, the capture tool and command, the channel/band setup, how many cycles, and a dmesg snippet showing RX actually stopping? #48 already makes the sniffer follow the channel on this chip, so I want to see the specific case it doesn't cover. As-is I'd also flag that the mac.c change drops the other DBDC chain's frames in monitor mode, which is a behavior change worth being deliberate about.

Ashcal9669:
Withdrawing this PR. I cannot provide a reproducible monitor-RX stop on top of the merged #48, and the current mt7925/mac.c frequency filter can discard frames from the other DBDC chain. The 7.0.12 failure used to motivate this patch was not isolated from module installation and loading state, so it is not sufficient evidence for these driver changes. Any future monitor change will be scoped to an exact reproduction with the capture command, channel and band setup, cycle count, and dmesg evidence.

Ashcal9669:
I reran this after closing the PR and reproduced a narrower failure on the exact PR base. The original lifecycle description was too broad.

Environment:

- MT7927 PCIe `14c3:7927`
- Linux `7.0.12+deb13-amd64`
- baseline commit `215a702528330795400b1a3940e753381204865a` (merged #48 and #49)
- out-of-tree `*_git` modules built directly from that commit
- existing interface converted to monitor mode

Setup:

```sh
make clean && make -j4

sudo ip link set wlp6s0 down
sudo iw dev wlp6s0 set type monitor
sudo iw reg set US
sudo ip link set wlp6s0 up
```

Reproduction sequence, with each command allowed to run for 10 seconds before SIGINT:

```sh
sudo timeout --signal INT 10s hcxdumptool -i wlp6s0 -c 44b --rcascan=active --rds=5
sudo timeout --signal INT 10s hcxdumptool -i wlp6s0 -c 11a --rcascan=active --rds=5
sudo timeout --signal INT 10s hcxdumptool -i wlp6s0 -c 44b --rcascan=active --rds=5
sudo timeout --signal INT 10s hcxdumptool -i wlp6s0 -c 11a --rcascan=active --rds=5
```

Baseline results:

| Run | Channel | Kernel packets | Beacons | Probe TX | Probe responses | Displayed APs |
| --- | --- | ---: | ---: | ---: | ---: | --- |
| 1 | 44 / 5 GHz | 3794 | 1665 | 2 | 41 | yes |
| 2 | 11 / 2.4 GHz | 3979 | 1726 | 0 | 0 | none |
| 3 | 44 / 5 GHz | 4378 | 1495 | 2 | 45 | yes |
| 4 | 11 / 2.4 GHz | 3977 | 1678 | 0 | 0 | none |

The 2.4 GHz runs still receive packets at the packet socket, but hcxdumptool displays no APs and transmits no probe requests. Switching back to 5 GHz immediately restores decoded APs and probe responses.

I then built commit `4a86e856ad4603843e80af1cf4cf4750aa665a4f` and ran the identical sequence:

| Run | Channel | Kernel packets | Beacons | Probe TX | Probe responses | Displayed APs |
| --- | --- | ---: | ---: | ---: | ---: | --- |
| 1 | 44 / 5 GHz | 4824 | 1590 | 2 | 38 | yes |
| 2 | 11 / 2.4 GHz | 1322 | 541 | 2 | 26 | yes |
| 3 | 44 / 5 GHz | 4719 | 1540 | 2 | 44 | yes |
| 4 | 11 / 2.4 GHz | 1144 | 488 | 2 | 39 | yes |

The dmesg delta for both runs contains only normal `entered promiscuous mode` / `left promiscuous mode` transitions, with no firmware reset, warning, or crash. The failure is visible in the capture behavior rather than a dmesg error.

This demonstrates a 5 GHz to 2.4 GHz sniffer transition that merged #48 does not fully cover on this system. I am keeping the PR closed for now because the `mt7925/mac.c` DBDC-frequency filter is unrelated to this reproduction and can discard the other chain's frames. That part should be removed before the minimal channel-transition change is reconsidered.

Lucid-Duck:
@Ashcal9669 Ran your active-scan sequence on my MT7927 and it reproduces cleanly. On the merged base the monitor stops entirely after a couple of channel swaps (runs 3 and 4 deliver zero packets); on 4a86e856 all 8 cycles stay healthy, 2.4 GHz probe requests and responses included. The channel-transition re-arm is a real fix.

Agreed the mac.c frequency filter should come out, for the reason you gave. If you resubmit just the main.c re-arm, I'll run this same A/B on it to confirm.

Ashcal9669:
@Lucid-Duck 

Resubmitted the standalone main.c channel-transition re-arm fix without the mac.c frequency filter here: [New PR 54]

PR #53, opened 2026-07-02
wifi: mt76: mt7925: stop advertising MT7927 STR fallback

Ashcal9669:
## Summary

- stop the MT7927 fallback from advertising simultaneous-link and frequency-separation capabilities that firmware does not report
- keep MT7927 MLO enabled through the existing chip-capability fallback
- leave the documented EML capability parsing and monitor-mode paths unchanged

## Root cause

The MT7927 fallback set `MAX_SIMUL_LINKS` and `FREQ_SEP_TYPE_IND` even though the firmware capability data does not report STR support. That caused MLO remain-on-channel setup to select the JOIN/DBDC request path. Firmware granted the first two-link session after reset, then stopped granting identical requests during immediate reconnects, leading to control-port failures and four-way handshake timeouts.

Keeping the default zero `MAX_SIMUL_LINKS` value selects the MLSR ROC request with automatic band selection while preserving MLO support.

## Validation

- built successfully against Linux `7.0.12+deb13-amd64`
- completed two immediate 90-second MLO sessions and one 60-second session after a monitor-mode transition
- every run reported `valid_links=0x3` with links on 2422 MHz and 5260 MHz
- both links installed GTK and IGTK keys
- no unexpected disconnects, control-port failures, four-way handshake timeouts, firmware resets, warnings, or crashes occurred in the run deltas
- monitor/injection regression: 993 packets on 2.4 GHz and 8,228 packets on 5 GHz with zero kernel drops; active probes received responses on both bands

This validates repeat two-link MLSR operation. It does not claim unsupported simultaneous STR operation.

### DIFF
--- a/mt7925/main.c
+++ b/mt7925/main.c
@@ -284,11 +284,6 @@ int mt7925_init_mlo_caps(struct mt792x_phy *phy)
 	ext_capab[0].eml_capabilities = phy->eml_cap;
 	ext_capab[0].mld_capa_and_ops =
 		u16_encode_bits(0, IEEE80211_MLD_CAP_OP_MAX_SIMUL_LINKS);
-	if (is_mt7927(phy->mt76->dev)) {
-		ext_capab[0].mld_capa_and_ops =
-			u16_encode_bits(1, IEEE80211_MLD_CAP_OP_MAX_SIMUL_LINKS) |
-			IEEE80211_MLD_CAP_OP_FREQ_SEP_TYPE_IND;
-	}
 
 	wiphy->flags |= WIPHY_FLAG_SUPPORTS_MLO;
 	wiphy->iftype_ext_capab = ext_capab;


Lucid-Duck:
@Ashcal9669 Been running this on my MT7927 today, and preliminary testing looks good so far, no regressions with the STR advertisement removed. I'd like to put it through more tomorrow before I call it definitive, but the early signs are positive. Will follow up.

Ashcal9669:
@Lucid-Duck 

Understood. For clarity, my testing on 7.0.12+deb13-amd64 verified monitor RX across 2.4/5/6 GHz and active probe injection on 2.4/5 GHz with zero kernel drops. PR #53 also restored repeat two-link MLSR MLO association without claiming unsupported STR operation. I’ll wait for your final test results.

Lucid-Duck:
@Ashcal9669 Ran a fuller A/B on my MT7927 (14c3:7927) against two 3-link MLDs, base vs the change (only those 6 lines), 3 fresh-reset rounds of 8 immediate reconnects each per driver.

On my controlled MLD, base drops immediate reconnects with a 4-way handshake timeout: 4 of 24 failed, 9 Reason: 15=4WAY_HANDSHAKE_TIMEOUT in dmesg. With the change, 0 of 24 and no timeouts. Second MLD (tri-band), same direction: base 6 of 24, change 0 of 24, though those failures weren't 4-way timeouts, so I'd call the first AP the actual reproduction and the second corroborating.

Initial association came up multi-link every time on both drivers, and the change holds full 3-link MLO (2.4 + 5 + 6, EHT 320MHz, ~3.8 Gbit/s), so it isn't falling back to a single link.

Monitor side is unaffected: RX fine on 2.4/5/6, active probe injection works on 2.4 and 5 the same as base, no drops or warnings in dmesg.

Net: base 10/48 immediate-reconnect failures, 0/48 with the change, no regression I can find. No objection from me, and if this goes up to linux-wireless I'm happy to add my Tested-by.

Ashcal9669:
Thanks. I opened the narrowed PR as the minimal mt7925/main.c-only re-arm change, with the unrelated mt7925/mac.c DBDC filter removed.

If this exact minimal version still matches the change you tested, may I carry your Tested-by for the upstream linux-wireless submission?

Please give me the exact line you want used.

Lucid-Duck:
@Ashcal9669 Confirmed. I checked the current #53 (f15c8f1c) and it is exactly what I tested: the removal of the is_mt7927 block in mt7925/main.c that advertised MAX_SIMUL_LINKS=1 and the FREQ_SEP_TYPE_IND flag. My A/B ran against that change, so the tag applies as it stands.

Please use:

```
Tested-by: Devin Wittmayer <lucid_duck@justthetip.ca>
```

Ashcal9669:
Thanks, Devin.

I've amended the commit to include your Tested-by tag.

I'll leave this PR for your review, @morrownr

PR #54, opened 2026-07-03
wifi: mt76: mt7925: re-arm MT7927 sniffer on channel changes

Ashcal9669:
## Summary

- re-arm the MT7927 sniffer from the active monitor channel context
- reconfigure the sniffer when monitor filtering is reapplied
- stop the sniffer before crossing MT7927 physical sniffer bands, then restart it on the new context

## Scope

This is the minimal `mt7925/main.c` re-arm change requested after the A/B testing on PR #51. It intentionally excludes the experimental `mt7925/mac.c` DBDC frequency filter because that filter can discard frames from the other RX chain. There are no Makefile or `mt7925/mcu.c` changes.

## Reason

Repeated monitor transitions between 5 GHz and 2.4 GHz can leave MT7927 receiving packets at the packet socket while hcxdumptool no longer decodes APs or transmits active probes. Re-arming from the active channel context keeps the sniffer configuration synchronized across those transitions.

## Validation

- the `mt7925/main.c` delta matches the re-arm logic tested at commit `4a86e856ad4603843e80af1cf4cf4750aa665a4f`
- `make -j$(nproc)` passes against Linux `7.0.12+deb13-amd64`
- Lucid-Duck's hardware A/B on MT7927 reproduced the failure on the merged base and kept all eight 5 GHz/2.4 GHz transition cycles healthy with this re-arm logic, including 2.4 GHz probe requests and responses

### DIFF
--- a/mt7925/main.c
+++ b/mt7925/main.c
@@ -789,6 +789,9 @@ mt7925_monitor_update_chan(struct mt792x_vif *mvif,
 	if (!chan)
 		return;
 
+	if (ctx)
+		mvif->phy->mt76->chandef = ctx->def;
+
 	mconf->mt76.band_idx = mt7927_band_idx(chan->band);
 	mconf->mt76.basic_rates_idx = MT792x_BASIC_RATES_TBL;
 	if (chan->band != NL80211_BAND_2GHZ)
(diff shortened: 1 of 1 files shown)


Lucid-Duck:
@Ashcal9669 Ran the same 8-cycle active-scan A/B on my MT7927, base vs 5a8dea3b. The minimal main.c-only version holds up: on the merged base the monitor dies after the second channel swap (runs 3 through 8 capture zero packets, and 2.4 GHz stops transmitting probes even in run 2), while on 5a8dea3b all eight cycles stay healthy on both bands, probe requests and responses included. Same result I got on #51, so dropping the mac.c filter cost nothing. The channel-transition re-arm is a real fix.

Tested-by if it goes up:

```
Tested-by: Devin Wittmayer <lucid_duck@justthetip.ca>
```

Ashcal9669:
Thanks, Devin.

I've amended the commit to include your Tested-by tag.

This PR is intentionally limited to the mt7925/main.c channel-transition re-arm discussed after PR #51. The unrelated mt7925/mac.c DBDC-frequency filter has been removed.

@morrownr, this is ready for your review.

Issue #52, opened 2026-07-01
Netgear A9000 (MT7925u) crashing OpenWRT


SufficentMuffin:
Hello, 

i just bought a Netgear A9000 WIFI 7 Stick with the plan to add an additional band to my router running on OpenWRT 25.12.2

I also noticed it also just shows up as "Generic MAC80211"

I am able to create an AP which is also seen by clients but as soon as I try to connect the router gets unresponsive and resarts after a few seconds. 

The running Kernel is 6.12.74 and I installed the following packages: 

kmod-mt7925-common	
kmod-mt7925-firmware		
kmod-mt7925u

Anything I missed here?

SufficentMuffin:
I want to add I also have an Brostrend Wifi Stick (MT7921au) which works perfectly fine. If that helps anything.

morrownr:
Hi @SufficentMuffin 

I have both of the adapters that you mention. Late last year I brinked my router that has USB ports. Maybe I need to unbrick it so I can test this. For now, I can just guess.

> I want to add I also have an Brostrend Wifi Stick (MT7921au) which works perfectly fine. If that helps anything.

Yes, that helps. Here is what I suspect: 25.12 is not fully ready for WiFi 7. You may want to try a development build. Here is where you can get them:

https://downloads.openwrt.org/snapshots/targets/

> The running Kernel is 6.12.74 and I installed the following packages:

You should just need to install kmod-mt7925u as the other needed files will be pulled automatically.

I have set my A9000 up in AP mode but it was not with openwrt. I use a Pi4B with RasPiOS to do testing so I used hostapd. This kind of a setup can get complicated once you get to WiFi7.

Lucid-Duck:
I spent today digging into this. Here is where it stands.

I can reproduce your crash on the same setup: OpenWrt 25.12.2, kernel 6.12.74, an A9000 as an AP on a Pi 5. The AP comes up fine, then the box goes unresponsive and reboots the moment a client finishes associating. So it is not your router or your config, it is the mt7925 driver on the AP side. Your mt7921au working as an AP fits, since it runs a different driver.

Updating will not help, since the same crash happens on the current snapshot. It also looks related to the older AP-mode report at openwrt/openwrt#16273.

[...]

PR #58, opened 2026-07-11
wifi: mt76: mt7925: ensure tx headroom in usb_sdio_tx_prepare_skb

Lucid-Duck:
This fixes the mt7925u AP crash in #52: an mt7925 USB adapter used as an AP reboots the machine when a connected client's traffic is forwarded.

The cause is a missing tx headroom reservation. `mt7925_usb_sdio_tx_prepare_skb()` prepends a TX descriptor plus a USB header (68 bytes) with `skb_push`. That headroom normally comes from `hw->extra_tx_headroom`, but forwarded data frames take the 802.3 encap-offload path (`ieee80211_8023_xmit`), which does not reserve it. The frame arrives 4 bytes short, `skb_push` underflows, and `skb_under_panic` reboots the box. Locally generated frames carry the slack and survive, so it shows up as a crash on the first forwarded frame after a client joins.

The fix grows the headroom with `skb_cow_head`, only for frames that are short, so it is a no-op on the common path.

Verified on hardware, Netgear A9000 (mt7925u) as a hostapd AP on a Pi 5 (RasPiOS 6.18.34):

- Stock: a bridged AP panics on client DHCP, 2 for 2. The captured trace matches the #52 reporter's oops on an Asus TUF AX4200 (OpenWrt 6.12.74), same function and the same 4-byte underflow.
- Patched: the same bridged AP takes association, DHCP, three reconnects and a forwarded ping flood with no panic.
- No regression: a matched same-session A/B on an isolated (non-forwarding) AP shows the same throughput stock versus patched.

The same code path is in the upstream mt7925 driver, so I will follow up with this to linux-wireless. Opening here first since the reporter and testers are on this repo.

### DIFF
--- a/mt7925/mac.c
+++ b/mt7925/mac.c
@@ -1435,6 +1435,10 @@ int mt7925_usb_sdio_tx_prepare_skb(struct mt76_dev *mdev, void *txwi_ptr,
 	if (!wcid)
 		wcid = &dev->mt76.global_wcid;
 
+	if (skb_headroom(skb) < MT_SDIO_TXD_SIZE + MT_SDIO_HDR_SIZE &&
+	    skb_cow_head(skb, MT_SDIO_TXD_SIZE + MT_SDIO_HDR_SIZE))
+		return -ENOMEM;
+
 	if (sta) {
 		struct mt792x_sta *msta = (struct mt792x_sta *)sta->drv_priv;
 


morrownr:
Merged. The more I think about this patch, the more I think it needs to go into 7.2 rc and be set to oackport.

Lucid-Duck:
@morrownr Following your go-ahead, I sent this upstream to linux-wireless as a `[PATCH wireless]` with `Cc: stable`: https://lore.kernel.org/linux-wireless/20260715023348.59506-1-lucid_duck@justthetip.ca/

One small difference from what merged here: the upstream version uses the exact mt7921 idiom (`err = skb_cow_head(...); if (err) return err;`) rather than the headroom-check form, functionally identical but consistent with the sibling driver. Happy to align the copy in this repo to match if you'd prefer.

The commit credits your Pi 4 control run, which is what pinned the trigger to the ingress path. The `Fixes:` tag points at the mt7925 driver-add, so once it lands it should backport across stable and reach OpenWrt on a later refresh. I'll follow up here if it needs a v2 or draws review comments.

PR #59, opened 2026-07-14
Add MAINTAINING.md

Lucid-Duck:
@morrownr,

This is the patch-handling write-up you asked for back in April, the "plan and document it for how to handle patches" one. I owe you an apology on the timing. I had a first draft done that same week and then it sat while the mt76 and other work piled up, and three months slipped by. That's on me, it should have been in your hands long ago.

Here it is now, and I wrote it around how you actually work: straight to main, revert if it breaks, no branch ceremony, with the send-me-the-diff option for anything that needs eyes first. It covers the openwrt cherry-pick steps, the compat guards, the build check, and what the compat-patches folder actually is. Every command and code example is checked against the current tree.

It's yours to change however you like. Merge it, edit it, or tell me what to fix. And if it overlaps the checklist you've been putting together, all the better, we can fold them into one.

Sorry again for sitting on it.

Devin

### DIFF
--- a/MAINTAINING.md
+++ b/MAINTAINING.md
@@ -1,0 +1,106 @@
+# Maintaining morrownr/mt76
+
+How patches get into this tree, written plain so anyone lending a hand can do it without breaking things. If something here doesn't match what you're actually seeing in the repo, trust the repo, not the doc.
+
+## What lands here, and where it comes from
+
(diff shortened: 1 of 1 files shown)


morrownr:
This is a good start and can be modified as needed. Thanks.

PR #61, opened 2026-07-22
feat(firmware): add MT7925 firmware and link to MT7927

Ashcal9669:
### Description
- **Update MT7925 BT Firmware**: Replaces the older firmware binary with the updated 119 KB version.
- **Link MT7927 BT Firmware**: Removes the stale `MT6639` binary from the `mt7927` directory and symlinks it to the updated `mt7925` firmware file since they are byte-for-byte identical.


Lucid-Duck (changes requested):
@morrownr this is the same commit as the first one in #66, so the same objection applies: https://github.com/morrownr/mt76/pull/66#pullrequestreview-4801284803

It deletes the Bluetooth firmware for the MT7927. That file is not in linux-firmware, and this repo is the only place I could find it. My own MT7927 had no working Bluetooth until I copied that exact file onto the system, and it has worked ever since, including across a reboot this afternoon. Deleting it would leave anyone relying on this repo's copy without Bluetooth.

It also swaps out the MT7925 Bluetooth file. The one in the repo today is the same file linux-firmware ships. The replacement is about a quarter of the size and its header does not look like MediaTek firmware. I could not test that one, as I have no MT7925 Bluetooth adapter, so I am going only on what is in the file.

The Makefile change in the same commit is a good fix and worth keeping. Without that flag, `make install` can produce compressed modules the kernel cannot unpack. Split into its own commit it could go straight in.

Marking changes requested for the firmware files.

morrownr:
@Ashcal9669 

I appreciate your work and want to encourage you to continue. I do owe you an explanation for why I am resisting this specific patch. The idea for this repo is to make things better but do it in a manner that is consistent with how mt76 is handled in mainline. This patch diverges from mainline in that it is using a bt firmware file that is not designed for mt7927.

This patch may work better right now but it includes things that would have to be undone to get back to where mainline is or will be when the real mt7927 bt firmware drops. Part of my job here is to help manage this repo and I am getting old. My ability to track numerous workarounds that will need to be cleaned up is not there anymore. I do the best that I can but I am a guy that learned programming with FORTRAN and COBOL back before we had keyboards, mice and monitors. Really! 

> The Makefile change in the same commit is a good fix and worth keeping. Without that flag, make install can produce compressed modules the kernel cannot unpack. Split into its own commit it could go straight in.

I am asking you to consider following Devin's (Lucid) guidance above. If that happens, the patch will go in asap.

I hope you understand and if you have questions, please let me know.

Nick

Ashcal9669:
@morrownr @Lucid-Duck Thanks both for the review. Force-pushed a rewrite of 245677eb addressing this:

- Split the `xz --check=crc32` Makefile fix into its own PR: #68 (unrelated to the firmware question, should be able to land independently).
- Firmware: 245677eb's file changes were a straight regression, not a fix -- both `firmware/mt7927/BT_RAM_CODE_MT6639_2_1_hdr.bin` and `firmware/mt7925/BT_RAM_CODE_MT7925_1_1_hdr.bin` were already correct in the repo before that commit. This PR now just drops those changes and restores the pre-regression files, rather than replacing them with anything new. Verified byte-for-byte against Lucid's findings: both carry the "ALPS" tag at offset 16 that the corrupted file was missing, and match the original sizes (688341 / 446831 bytes) exactly. Also fixed the README's documented install path for the MT7927 BT file, which was wrong independent of 245677eb.

@morrownr on the mainline-consistency concern: understood, and to be clear this isn't introducing a new non-mainline workaround -- it's reverting a regression back to firmware that was already sitting in the repo prior to 245677eb. Nothing new is being added here that wasn't already present.

Lucid-Duck (approved):
Cleared, and thanks for splitting it.

This is only the README path fix now, and the correction is right. btusb derives the path from the chip id and asks for mediatek/mt7927/BT_RAM_CODE_MT6639_2_1_hdr.bin. That is exactly where I installed the file on my own 7927 today, and it is what got Bluetooth working on it. It is also the only BT path in that README missing the mediatek/ component; the other five are already correct.

PR #66, opened 2026-07-27
wifi: mt76: mt7925: re-arm MT7927 sniffer on assign_vif_chanctx

Ashcal9669:
mt7925_assign_vif_chanctx never re-armed the firmware sniffer, so a
monitor vif stayed stuck on whatever band it was first enabled on:
change_chanctx re-armed correctly, but iw's cross-band channel set on
an unassigned monitor vif goes through assign_vif_chanctx instead,
which only updated the chanctx pointer.

Factor the re-arm sequence (update_chan + set_sniffer(true) +
config_sniffer + clear beacon filter) into a shared
mt7925_monitor_arm_sniffer() helper and call it from both chanctx
paths. Also keeps the earlier beacon-filter guards: a monitor vif must
never get IEEE80211_VIF_BEACON_FILTER or PM applied, or firmware drops
other-BSS beacons.

Verified on MT7927: full 2.4->5->6->5->2.4 GHz band round-trip on a
single monitor vif with no interface recreation, beacons received on
every hop, no firmware warning/reset in dmesg. Packet injection via
aireplay-ng --test against an owned AP: 29/30 (96%) on 2.4GHz ch11,
30/30 (100%) on 5GHz DFS ch64.


Ashcal9669:
Follow-up commit fixes a correctness bug in the original patch: the cross-band teardown call to \`mt7925_mcu_set_sniffer(dev, vif, false)\` was silently reporting the *new* band_idx instead of the old one, because PR #65 changed that function to resolve band_idx from the vif's own chanctx pointer rather than the PHY chandef, and that pointer is already-updated (or already the mutated-in-place object) by the time this runs. Fixed by temporarily clearing the ctx pointer for that one call so it falls back to the still-stale PHY chandef. No hardware regression was observed from the original bug (single monitor vif can only be tuned to one band at a time), but the teardown call wasn't doing what its own comment claimed.

Lucid-Duck (commented):
@morrownr thanks for the review request.

**`245677eb` should not merge.** It removes `firmware/mt7927/BT_RAM_CODE_MT6639_2_1_hdr.bin`. On my 7927 that is the exact path btusb asks for: the log line is `Direct firmware load for mediatek/mt7927/BT_RAM_CODE_MT6639_2_1_hdr.bin`, about 170 failures a minute, every boot. linux-firmware ships no MT7927 BT blob. Putting this repo's copy into `/lib/firmware/updates/` stopped the loop and brought Bluetooth up, and the driver logged `Build Time: 20250606201235`, which is the header of that exact file. The symlink the commit adds is named `BT_RAM_CODE_MT7925_1_1_hdr.bin`, which is not a path my chip requests.

The same commit also replaces `firmware/mt7925/BT_RAM_CODE_MT7925_1_1_hdr.bin`. The current file is byte-identical to linux-firmware's at 446831 bytes; the replacement is 119328 bytes, carries no `ALPS` platform tag at offset 16, and its `section_num` reads 3521693024 where the genuine 7925 and 7927 blobs read 5 and 9. To be clear about the limit of that one: it is header inspection only. I have no MT7925 Bluetooth device here, so I have not loaded it and cannot say what it does on hardware.

**The two code commits: no regression, but I could not reproduce the problem.** Stock `905c0d79` against the same plus this PR merged, so the only delta is the Makefile line and the 80 lines of `mt7925/main.c`. One monitor vif walking bands, beacons counted per hop. On my 7927, stock alone retuned on every cross-band hop including 6 GHz, so there was nothing to recover. On an MT7925 USB adapter I ran both builds and the counts are indistinguishable, though 6 GHz was refused on that adapter on every build including the in-kernel one, so I could not cover that band there.

ftrace on both chips shows `mt7925_change_chanctx` firing about a microsecond after `mt7925_assign_vif_chanctx` on every `iw set freq`, cross-band included. Since the change path already re-arms the sniffer, the missing re-arm in the assign path has no effect I can measure. That may be the difference from the run described in the commit message.

I have not run the patched build on the 7927 itself. `rmmod` hung in stock main's napi teardown before I could swap, which is not this PR. Raising that separately.

I'll be testing this more later today when I get the time.

Lucid-Duck (changes requested):
Marking this as changes requested, for the firmware commit `245677eb` specifically. Detail in my review above. The two code commits are not the issue.

Ashcal9669:
@Lucid-Duck @morrownr Rewrote the commit message (force-pushed, code content unchanged from the last push) to fix the actual justification rather than argue with the ftrace finding, which I've verified is correct.

**Lucid's finding stands and I confirmed the mechanism**: in `net/mac80211/chan.c`, `ieee80211_assign_link_chanctx()` unconditionally calls `ieee80211_recalc_chanctx_min_def()` right after `assign_vif_chanctx`, which fires `change_chanctx` whenever anything differs -- reliably true for a solo monitor vif retuning to a fresh chanctx. So on the `iw set freq` path, `change_chanctx`'s existing re-arm (PR #54) does the real work, exactly as your ftrace showed. That part of the original commit message was wrong and I'm not defending it.

**Where this fix actually matters**: `ieee80211_reconfig()` (`util.c:2038-2040`) -- run after a firmware-recovery restart or resume from suspend -- reassigns the monitor vif's chanctx via `drv_assign_vif_chanctx()` directly, with no recalc and no `change_chanctx` follow-up at all. I reproduced the underlying symptom on hardware: monitor vif on a non-default 5GHz channel, triggered a firmware `chip_reset`, and beacon reception silently dropped to near zero afterward even though mac80211 still reported the correct channel. That's this path failing with no safety net.

Caveat on the hardware repro: I tested against this project's separate hardware-validation driver tree (its own more complex chanctx-lifecycle rewrite), not this exact diff, since I'm keeping that tree and this repo separate. That tree still exhibits the symptom, for what looks like a different, more specific reason (an early bailout when a WCID isn't populated yet post-reset) that this simpler patch's unconditional re-arm call doesn't share. So: the *problem* (reconfig path has no re-arm safety net) is hardware-confirmed; this patch's fix for it is verified by reading assign_vif_chanctx is the same driver callback in both paths, not by running this exact diff through a live chip_reset.

Also confirmed while in here: PR #65's CNM-enable commit for MT7927 (Sean Wang, already merged) is what makes assign_vif_chanctx/change_chanctx real driver callbacks here instead of mac80211's single-channel emulation fallback -- this fix depends on that being in place, and it is.

Lucid-Duck (changes requested):
@Ashcal9669 thanks for digging into this and finding the reconfig path. That is a much stronger justification.

I checked it and it holds up. In current mainline, ieee80211_assign_chanctx() (util.c:1712) calls drv_assign_vif_chanctx() and then returns, with nothing after it, and ieee80211_reconfig() is the only caller, twice, at 2043 and 2104. So after a firmware recovery or a resume there is nothing to re-arm the sniffer. That is a real gap and this closes it.

On the iw path, for what it is worth. On an MT7925 USB adapter I ran stock 905c0d79 and the same plus this PR, one monitor vif walking 2.4 and 5 GHz with beacons counted per hop, and the counts track each other within normal beacon variance. On the MT7927 itself I have only got stock through so far, where every cross-band hop retuned correctly including 6 GHz. So no regression that I can see.

The one thing still holding this up is the firmware commit 245677eb, which is unchanged in here. #61 shows what it looks like without it. If you drop it from this branch I will clear this straight away, since the two code commits are not the problem.

Ashcal9669:
@Lucid-Duck Thanks for verifying the util.c mechanism independently -- good to have that confirmed against mainline directly rather than just this fork's history.

Rebased this branch onto current main (post-#61) to drop 245677eb from its ancestry, force-pushed. Diff is now just the two driver commits against mt7925/main.c -- no firmware files. Verified locally before pushing: `245677eb` is no longer an ancestor of this branch, build is clean, and the re-arm code is unchanged content-wise from before the rebase.

Lucid-Duck (approved):
@Ashcal9669 Confirmed on MT7927 at 991d9d0. Same-base A/B against 8da5cdc4, one monitor vif, beacon count before and after ieee80211_reconfig with no retune in between. Triggered through mac80211's debugfs reset so the host never suspends.

| band | stock | patched |
|---|---|---|
| 2437 | 933 -> 1 | 915 -> 892 |
| 5180 | 137 -> 1 | 137 -> 139 |
| 6135 | 138 -> 1 | 137 -> 137 |

Stock reports the correct channel the whole time while firmware is deaf. ftrace shows the helper firing off assign_vif_chanctx on patched and nothing on stock.

Two things. All three bands fail the same way on stock, 2.4 included, so it does not look band-conditional. And chip_reset looks like a separate problem: mt7925_mac_reset_work never calls ieee80211_restart_hw, so reconfig is never reached and both builds drop to 1 beacon there. Might be worth a follow-up.

Clearing my changes-requested, the firmware commit is gone from the branch.

Ashcal9669:
@Lucid-Duck Appreciate the real stock-vs-patched A/B, that's much better evidence than my chip_reset repro was.

Worth correcting on my end: my earlier hardware test used chip_reset and I attributed the beacon drop to the ieee80211_reconfig() gap this PR fixes. Your finding that mt7925_mac_reset_work never calls ieee80211_restart_hw() means chip_reset doesn't reach reconfig() at all -- so that repro showed a real symptom, just not proof of *this* mechanism. Your test (beacon count across an actual reconfig, band-independent per your data) is the real validation here.

Agreed the chip_reset/mac_reset_work path not reaching ieee80211_restart_hw is a separate bug worth its own follow-up -- happy to help dig into that one too if useful. Thanks for catching both.

PR #67, opened 2026-07-28
wifi: mt76: mt7925 + mt7615: drop TXRX_NOTIFY on non-mmio buses

Lucid-Duck:
The two TXRX_NOTIFY guards from the #32 list, ported as offered. mt7921 already came in through PR #7, so this completes the set: upstream these three were one series, patches 2, 3 and 4 of `20260627191336.20223`, and mainline now carries all of them as `da4082e91aca`, `feeff151c83e` and `39afc46c0243`.

What they fix: `PKT_TYPE_TXRX_NOTIFY` is an mmio-only event, but `mt7925_rx_check()` / `mt7925_queue_rx_skb()` and their mt7615 counterparts dispatch it to `*_mac_tx_free()` on every bus. That path cleans the DMA tx queues through `queue_ops->tx_cleanup()`, which only the mmio queue ops implement. On USB and SDIO it is NULL, so the event calls a NULL pointer from the RX worker. Both commits drop the event via `mt76_is_mmio()`, the same shape as Lorenzo's `5683e1488aa9`. In this tree that reaches `mt7925u_git.ko` and `mt7663u_git.ko` / `mt7663s_git.ko`.

The mt7925 case label has been there since the driver landed (`2afc7285` here, `c948b5da6bbe` upstream), which is why this looks like a no-op at a glance. It is not: the label existed, the bus check never did.

Cherry-picked with `-x`, authorship and the upstream `Signed-off-by`/`Link` trailers preserved. Both diffs are line-identical to the mainline commits. Build tested on 7.1.5 against today's main after your cherry-pick wave: clean rebase, no conflicts, full `make clean` then full build, 26 modules. checkpatch `--strict` clean on both.

### DIFF
--- a/mt7615/mac.c
+++ b/mt7615/mac.c
@@ -1601,6 +1601,8 @@ bool mt7615_rx_check(struct mt76_dev *mdev, void *data, int len)
 
 	switch (type) {
 	case PKT_TYPE_TXRX_NOTIFY:
+		if (!mt76_is_mmio(mdev))
+			return false;
 		mt7615_mac_tx_free(dev, data, len);
 		return false;
 	case PKT_TYPE_TXS:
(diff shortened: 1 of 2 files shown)


morrownr:
@Lucid-Duck 

Thank you very much.

PR #68, opened 2026-07-28
Makefile: use xz --check=crc32 for compressed module install

Ashcal9669:
Split out of #61 per Nick's request — this half is independent of the firmware discussion there and can land on its own.

Some kernels build their in-tree xz-compressed modules with the crc32 integrity check instead of xz's default (crc64), and their module loader only accepts the check type matching the kernel's own scheme. Compressing our out-of-tree modules with the default xz check produces a .ko.xz the kernel's decompressor refuses to unpack on those kernels, making `make install` silently produce modules that never load.

Match the distro's own compression scheme by using --check=crc32.


morrownr:
Awesome!

PR #69, opened 2026-07-29
wifi: mt76: mt7925: re-arm monitor sniffer on chip_reset recovery

Ashcal9669:
Follow-up to #66, per Devin Wittmayer's finding there: `chip_reset` drops monitor beacon reception to ~1/6s on both stock and patched builds because `mt7925_mac_reset_work()` never reaches `ieee80211_restart_hw()`/`ieee80211_reconfig()` at all — this is a separate bug in the driver's own reset-recovery path, not the mac80211-level gap #66 fixes.

`mt7925_mac_reset_work()`'s recovery reconnects STATION/AP vifs via `mt7925_vif_connect_iter()`, which has no monitor-vif handling whatsoever — it never calls `mt7925_mcu_set_sniffer()`/`mt7925_mcu_config_sniffer()` for one. mac80211's logical state (interface up, correct channel) survives the reset; firmware's sniffer state does not, and nothing puts it back.

Fix reuses the existing `mt7925_sniffer_interface_iter()` helper — the same mechanism `mt7925_config()` (on `IEEE80211_CONF_CHANGE_MONITOR`) and `mt7925_configure_filter()` already use — rather than adding new logic. Exposed the previously-static helper via `mt7925.h` for `mac.c` to call.

Not yet hardware-tested against a live `chip_reset` — flagging that plainly rather than claiming validation I don't have. Would appreciate an A/B check similar to what was done for #66 if anyone has hardware handy.


Ashcal9669:
Force-pushed a correction to this patch -- the original version had two real problems I should have caught before opening this:

1. It reused `mt7925_sniffer_interface_iter()` unscoped, applying `mt7925_mcu_set_sniffer(dev, vif, true)` to every active interface whenever a monitor vif happened to be up (the function only checks a hw-level monitor flag, not per-vif type). That would incorrectly tell firmware to treat a STATION/AP vif as a sniffer too if one coexists with a monitor vif during reset recovery.
2. None of the underlying MCU calls (`set_sniffer`, `config_sniffer`, `set_beacon_filter`, `set_deep_sleep`) had their return values checked or propagated.

Fixed: added a `struct mt7925_sniffer_iter_data` (dev/error/monitor_only) so the iterator can restrict itself to actual monitor vifs and propagate every MCU call's failure, plus a dedicated `mt7925_sniffer_rearm()` entry point (monitor-only) that the reset path calls under `dev->mt76.mutex`, with the error logged and queues woken only after the re-arm attempt completes. `mt7925_config()`/`mt7925_configure_filter()`'s existing call sites now use the same struct (unscoped, same behavior as before) and log the iterator's error instead of silently ignoring it.

Verified: clean build, and added a structural contract (`tests/mt7927/reset-sniffer-recovery-contract.sh`, not part of this PR) checking for the error-carrying state, monitor-only restriction, mutex serialization, failure logging, and correct re-arm-before-wake ordering.

Still not hardware-tested against a live chip_reset, same caveat as before.

Lucid-Duck:
@Ashcal9669 Tested it. It does not re-arm on either chip here, for two different reasons.

MT7927 on 7.1.5: mt7925_sniffer_rearm never runs. mt792x_core.c sets NO_VIRTUAL_MONITOR when is_mt7927(), and mac80211 only sets IEEE80211_CONF_MONITOR in the branch that flag skips, so the guard in mac.c is never true on this chip. ftrace shows mac_reset_work and nothing after it.

MT7925 on 7.2-rc5: the re-arm does fire, but beacons still go to 1. The iterator keeps the is_mt7927 gate on config_sniffer, so the sniffer is enabled without ever being configured for the channel. change_chanctx calls config_sniffer unconditionally for a monitor vif, which is why that path works.

| band | stock | #69 |
|---|---|---|
| 2437 | 205 -> 1 | 204 -> 1 |
| 5180 | 135 -> 1 | 136 -> 1 |

Keying off vif->type rather than the hw flag, and dropping the is_mt7927 gate on config_sniffer, makes it work on both here.

Ashcal9669:
@Lucid-Duck Thanks for the real hardware A/B, that pinpointed two distinct bugs I missed:

1. **MT7927**: the reset-recovery guard checked `hw->conf.flags & IEEE80211_CONF_MONITOR`, but `mt792x_core.c` sets `NO_VIRTUAL_MONITOR` for MT7927, and mac80211 only sets that flag in the branch `NO_VIRTUAL_MONITOR` causes it to skip -- so `mt7925_sniffer_rearm()` never ran on this chip at all.
2. **MT7925**: the re-arm did fire, but `mt7925_mcu_config_sniffer()` was still gated behind `is_mt7927()` (inherited from the iterator's original `mt7925_configure_filter()` caller, which is MT7927-only), so sniffer got enabled without ever being told the channel.

Force-pushed: monitor state now derives from `vif->type == NL80211_IFTYPE_MONITOR` instead of the hw flag, and dropped the `is_mt7927()` gate on `config_sniffer` so it runs for any monitor vif on either chip -- matching what `change_chanctx` already does unconditionally.

Not able to re-verify against a live chip_reset myself right now -- would appreciate the same stock-vs-patched check you ran before if you have hardware time for it.

Ashcal9669:
Force-pushed another fix on top -- real hardware testing (chip_reset + beacon counting on both chips, not just the structural contract) showed the previous commit still didn't work. The reason: the reset-work caller in `mac.c` still gated the call to `mt7925_sniffer_rearm()` behind `hw->conf.flags & IEEE80211_CONF_MONITOR`, the exact flag that's permanently false on MT7927. Fixing the iterator's internal logic didn't matter if the function was never being called at all. Now calls it unconditionally (it already scopes itself to actual monitor vifs, so this is a safe no-op with none active).

Lucid-Duck:
@Ashcal9669 Worth holding this one. The re-arm call in mac.c takes dev->mutex across ieee80211_iterate_active_interfaces(), which takes local->iflist_mtx. mac80211 already goes the other way, holding iflist_mtx and calling driver ops that take dev->mutex, via ieee80211_del_virtual_monitor -> drv_unassign_vif_chanctx -> mt792x_unassign_vif_chanctx. That is an ABBA inversion.

I built a 7.2-rc5 kernel with PROVE_LOCKING and confirmed it on MT7922:

```
-> #1 (&local->iflist_mtx):  ieee80211_iterate_interfaces <- mt7921_config
-> #0 (&dev->mutex#3):       mt792x_unassign_vif_chanctx <- ieee80211_del_virtual_monitor
*** DEADLOCK ***
```

Taking the lock inside the iterator callback instead, so the order is only ever iflist_mtx then dev->mutex, is the right direction.

Fair warning it does not end there. The driver has roughly a dozen sites taking dev->mutex across ieee80211_iterate_*, including mt7921_config and mt7925_config, and I get the same splat on completely stock code with no patch applied. So this is pre-existing and your patch inherits it rather than causes it. I am writing that up for linux-wireless separately.

Lucid-Duck:
@Ashcal9669

Ash, this one is unstuck, though not the way I expected.

The inversion has a fix upstream now and Felix took it, but it does not reach your call site. The config paths get away with the mutex-asserting iterator because mac80211 holds the wiphy mutex on the way in, and that variant checks for it. The reset worker is a plain work item with nothing held, so the same swap would trip the assert instead of fixing anything. Took me until today to see why.

The order is the thing. Take the device mutex inside the callback rather than around the walk, and it only ever goes list lock then device mutex, which is where mac80211 was already headed.

Your force-push on the 30th is already the right shape for it. The iterator data struct and the monitor-only entry point are exactly what this needs.

PR #70, opened 2026-07-30
wifi: mt76: mt7925: fix rmmod hang from redundant napi_disable

Ashcal9669:
Found while testing PR #69 on real hardware: `rmmod mt7925e_git` hung indefinitely, confirmed stuck via the blocked task's kernel stack:

\`\`\`
napi_disable_locked
napi_disable
mt76_dma_cleanup [mt76_git]
mt7925_pci_remove [mt7925e_git]
\`\`\`

Two bugs combine to cause it:

1. `mt7925e_unregister_device()` calls `napi_disable()` on every RX queue directly, then `mt792x_dma_cleanup()` (via `mt76_dma_cleanup()`) calls `napi_disable()` on the *same* queues again moments later before properly deleting them. Two unpaired disables on one NAPI struct with no `napi_enable()` between them -- the first is redundant.
2. `mt792x_irq_tasklet()` never checks `MT76_REMOVED` before touching hardware registers or scheduling NAPI, unlike every other bus type in this driver family (`usb.c`, `mt76x02_usb_mcu.c` already do). If the tasklet runs once more between the two `napi_disable()` calls above and reschedules NAPI, the second, unpaired disable has nothing left to wait for.

**Verification status, plainly:** clean build only. I have not confirmed this actually prevents the hang on real hardware -- I hit this exact hang mid-test and the box needs a reboot to recover before I can retest the actual `rmmod` scenario, which I haven't done yet. Flagging that honestly rather than claiming it's fixed.


Lucid-Duck:
@Ashcal9669 Detail on the #32 note, since it's your PR.

The race Sashiko found: `PKT_TYPE_TXRX_NOTIFY` reaches `mt76_token_release()` and `idr_remove()` while `mt7925_tx_token_put()` runs `idr_destroy()` a few lines below. The loops you're removing were what kept those apart. Mikhail has offered a v2 if Nicolas doesn't want it himself: revert `13b7e6a96a00`, then disable napi in mt7915's own unregister path.

On your tasklet check, I looked: the only `tasklet_kill()` in either `mt7921/pci.c` or `mt7925/pci.c` is in the suspend path, so remove really does go `tasklet_disable()` then `mt76_free_device()` with nothing in between. That half stands on its own.

Thanks for all the work you've put in here. Not many people have this level of technical ability, and the repo is better for having it.

Ashcal9669:
Reopened as #71 with the race fix from the review above (napi_disable()/napi_enable() restored around mt7925_tx_token_put(), plus a driver-ownership fix found in additional review). Couldn't reopen this PR directly — GitHub blocks reopening once the branch history has been force-pushed.

PR #73, opened 2026-08-01
wifi: mt76: harden mt7925 remove and firmware recovery paths

Ashcal9669:
Supersedes #71 on top of current `main` (which already contains #72).

This carries the current local recovery/remove-state fixes:

- skip connac/mt7925 MCU BSS/dev/chctx updates once `MT76_REMOVED` is set
- make thermal/mac-work paths return cleanly after remove
- disable only allocated RX NAPI queues in `mt7925e_mac_reset()`
- unwind the `mt792xe_mcu_fw_pmctrl()` failure path through `out:`
- add an MT7925-specific firmware-download recovery path that resets WFSYS and restarts MCU if firmware does not reach `FW_PWR_ON`

Verification run on this local tree:

- `sh tests/mt7927/connac-add-dev-removed-disable-contract.sh /home/pi/mt76`
- `sh tests/mt7927/thermal-removed-no-mcu-contract.sh /home/pi/mt76`
- `sh tests/mt7927/removed-skips-bss-update-contract.sh /home/pi/mt76`
- `sh tests/mt7927/reset-error-unwind-contract.sh /home/pi/mt76`
- `sh tests/mt7927/reset-napi-allocated-queues-contract.sh /home/pi/mt76`
- `sh tests/mt7927/mac-work-removed-requeue-contract.sh /home/pi/mt76`
- `git diff --check`
- `make clean && make -j"$(nproc)" KVER="$(uname -r)"`

This PR is intentionally based on the current local branch state after #72, not on the earlier stale branch snapshots.

PR #74, opened 2026-08-01
wifi: mt76: mt7925: post-#72 teardown fix follow-up

Ashcal9669:
Follow-up to #71.

After #72 reverted the shared second `napi_disable()`, this branch carries the current MT7925 teardown fix candidate on a restored PR #73 so review can happen against the right thread.

This version keeps the pieces that still matter after #72:
- `MT76_REMOVED` early exits in `mt792x_irq_tasklet()`, `mt792x_poll_tx()`, and `mt792x_poll_rx()`
- RX NAPI quiesced before `mt7925_tx_token_put()` so `idr_destroy()` does not race an in-flight `mt76_token_release()` path
- `tasklet_kill(&dev->mt76.irq_tasklet)` in `mt7925_pci_remove()` to close the remove-time tasklet lifetime gap
- the extra pre-unregister `__mt792x_mcu_drv_pmctrl(dev)` needed so `mt76_unregister_device()` teardown runs with driver ownership

Reference review context:
- #71 contains the detailed teardown/race analysis
- #72 reverted the shared second disable, which changed the final shape of the MT7925-side fix

Current goal for this PR: review the actual post-#72 MT7925 fix in its own thread without changing the already-open #71 history.

### DIFF
--- a/mt7925/pci.c
+++ b/mt7925/pci.c
@@ -40,29 +40,64 @@ static int mt7925e_init_reset(struct mt792x_dev *dev)
 
 static void mt7925e_unregister_device(struct mt792x_dev *dev)
 {
-	int i;
 	struct mt76_connac_pm *pm = &dev->pm;
 	struct ieee80211_hw *hw = mt76_hw(dev);
+	int i;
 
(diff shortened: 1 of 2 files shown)


Ashcal9669:
Closing this superseded thread. The focused post-#72 fix is now in #75: https://github.com/morrownr/mt76/pull/75

PR #75, opened 2026-08-01
wifi: mt76: mt7925: fix remove-time tasklet race

Ashcal9669:
This is the post-#72 replacement for the #70/#71 teardown discussion.

#72 removed the shared RX `napi_disable()` from `mt76_dma_cleanup()`, so this patch leaves `mt7925e_unregister_device()`'s existing driver-side RX NAPI quiesce before `mt7925_tx_token_put()` intact and does not re-enable RX NAPI before queue teardown.

Fixes included:

- add `MT76_REMOVED` exits to `mt792x_irq_tasklet()`, `mt792x_poll_tx()`, and `mt792x_poll_rx()`
- remove the unregister-side `tasklet_disable()` from `mt7925e_unregister_device()`
- add `tasklet_kill(&dev->mt76.irq_tasklet)` in `mt7925_pci_remove()` after `devm_free_irq()` and before `mt76_free_device()`

This closes the remaining remove-time tasklet/NAPI lifetime race without carrying the broader #71 pmctrl/NAPI restructuring.

Validation:

- `PR75 source contract: PASS`
- `git diff --check`: PASS
- `checkpatch.pl --strict --no-tree`: 0 errors, 0 warnings
- `make clean && make -j"$(nproc)" KVER="$(uname -r)"`: PASS on `7.2.0-rc1-mt7927-monitor-v3`

References:

- #70
- #71
- #72
- https://lore.kernel.org/all/CABXGCsO07SExb+Z0PeN6MZ1fKC24Tvn3ehSyeQc-3qFC7jM7dQ@mail.gmail.com/

### DIFF
--- a/mt7925/pci.c
+++ b/mt7925/pci.c
@@ -61,8 +61,6 @@ static void mt7925e_unregister_device(struct mt792x_dev *dev)
 	mt792x_dma_cleanup(dev);
 	mt792x_wfsys_reset(dev);
 	skb_queue_purge(&dev->mt76.mcu.res_q);
-
-	tasklet_disable(&dev->mt76.irq_tasklet);
 }
 
 static void mt7925_reg_remap_restore(struct mt792x_dev *dev)
@@ -725,6 +723,7 @@ static void mt7925_pci_remove(struct pci_dev *pdev)
 	set_bit(MT76_REMOVED, &mdev->phy.state);
 	mt7925e_unregister_device(dev);
 	devm_free_irq(&pdev->dev, pdev->irq, dev);
+	tasklet_kill(&dev->mt76.irq_tasklet);
 	mt76_free_device(&dev->mt76);
 	pci_free_irq_vectors(pdev);
 }
--- a/mt792x_dma.c
+++ b/mt792x_dma.c
@@ -31,6 +31,9 @@ void mt792x_irq_tasklet(unsigned long data)
 	const struct mt792x_irq_map *irq_map = dev->irq_map;
 	u32 intr, mask = 0;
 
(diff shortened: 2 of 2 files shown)


Lucid-Duck:
@Ashcal9669 Ran the traffic test from #71 against this branch: 10 of 10 rmmod cycles on my MT7927 with the vif associated and roughly 700 pps of receive traffic in flight at remove time, no hang, no spin, no warnings, and an identical 10 of 10 control run on current main.

One thing surfaced, on main and this branch equally: with a live vif, remove takes about a minute. Every teardown MCU command times out, because MT76_REMOVED is set before unregister and mt792x_irq_handler() returns IRQ_NONE once the bit is up, so command responses stop being delivered. Idle unloads never show it because there is almost nothing to tear down. Mainline sets the bit after unregister and does not have this.

Merging this. #71 can close, or narrow to the pmctrl question if you want to keep making that case. The set_bit ordering is a clean follow-up if you want it.

Thanks for the persistence on this one. Carrying a fix through this many rounds is real work, and the driver is better for it.

Ashcal9669:
@Lucid-Duck @morrownr 
Thanks. I see #75 is merged and scoped only to the remove-time tasklet/NAPI lifetime fix.

 I’m keeping the active-monitor work separate. The next focus is the `mt792x` active monitor path that Felix disabled for 7.3 with `wifi: mt76: mt792x: do not advertise active monitor`.

Current target: make active monitor work correctly again instead of just advertising it or disabling it. I’m going to focus on MT7925 first, then regression-check MT7927 so the MT7927 path is not broken while restoring active monitor support.

I’ll avoid mixing this with #75/#71 teardown work. If I find a real fix, I’ll submit it as a separate focused PR with evidence.

PR #77, opened 2026-08-04
MAINTAINING.md: document deliberate differences from upstream

Ashcal9669:
Per Lucid-Duck's note in #32: two spots in the tree are intentionally not what openwrt/mt76 or Felix's tree does -- the reverted double napi_disable() (#72) and mt7925's unconditional NO_VIRTUAL_MONITOR (#76). Neither was written down anywhere. This adds a short section to MAINTAINING.md next to the cherry-pick guidance, explaining each one and what to check if a future cherry-pick touches that area.

Documentation only, no driver code changed.

### DIFF
--- a/MAINTAINING.md
+++ b/MAINTAINING.md
@@ -71,6 +71,14 @@ Short commit id or long, git takes any unique piece of it. Use the 12-character 
 
 When a cherry-pick doesn't apply cleanly, it's almost always one of two shapes. Either openwrt changed a function we've wrapped in a compat guard, in which case keep their change and put the guard back around the new version. Or openwrt touched code we'd already patched a different way, which is rarer, and you look at both and decide what to keep.
 
+## Deliberate differences from upstream
+
+Two spots in the tree are intentionally not what openwrt/mt76 (or Felix's tree) does. Both were changed on purpose, for a reason written down in the commit, and neither should get silently overwritten by a future cherry-pick.
+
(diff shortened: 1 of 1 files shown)


Lucid-Duck:
@Ashcal9669 The idea here is right and Nick's asked for it, so this should definitely land. Three things I'd fix first, all in the wording rather than the approach.

The heading says mt7925, but the change went into mt792x_core.c, so it covers mt7921 and mt7922 as well. The body does mention the mt7921 testing, it's just the heading that's narrow, and someone hitting a conflict on mt7921 would reasonably skim it and think it isn't theirs. That's the one accident this section exists to prevent.

8be40d4b194c is our SHA rather than upstream's. Upstream is 332bbe9b2784 in openwrt and 13b7e6a96a00 in mainline. Anyone following the doc during a pick would go looking for a commit that isn't there.

And the napi one probably isn't a permanent difference. Mikhail posted the same revert upstream yesterday, so once it lands we're back in sync. Worth marking that entry as temporary so nobody's confused when the trees converge. The mt7915 note in there is genuinely useful though, keep that.

Lucid-Duck:
@Ashcal9669 Merged as #95, with your authorship on the commit.

Three changes, all from the August review. The section sits below the skip list now, which is what the conflict was. The commit id is openwrt's and mainline's rather than this tree's own. And the second heading says mt792x, since the flag is set in `mt792x_core.c` and covers mt7921 and mt7922 as well.

Nothing else changed. I went through the rest against the trees this morning and it holds, the mt7915 note included.

One thing worth knowing for later: Mikhail's revert reached 7.2 and Felix has it queued, so the napi entry has an end date on it. openwrt still carries the commit, which is why the warning earns its place for now.

Thanks for writing it.

PR #84, opened 2026-08-20
Rename picks-skip.txt to skip-picks.txt, and add a section for pending patches

Lucid-Duck:
@morrownr Both things from your mail.

Two commits. The first is the rename on its own, so git records it as a rename and `git log --follow skip-picks.txt` still traces back through your entries commit. It also updates the three references in `show-picks.sh` and the one in `MAINTAINING.md`.

The second appends your section below the existing list, in your wording. I added one line under it: keep those entries commented. `show-picks.sh` treats any uncommented token as an id to skip, so a bare id there would be counted and the "Skipping N commit(s)" line would come out wrong.

Ran the script afterwards and it reports `Skipping 3 commit(s) listed in skip-picks.txt.`


morrownr:
Thank you so much. I know if seemed like a little detail but I think this will work out slightly better. We will see.

PR #85, opened 2026-08-21
wifi: mt76: mt7925: disable firmware-level frame drop in monitor mode

Ashcal9669:
## Summary

`mt7925_mcu_sniffer_interface_update()` (mt7925/mcu.c) sets `drop_err = 1`
unconditionally in the `UNI_SNIFFER_CONFIG` MCU command. This tells the
firmware to filter weak/marginal management frames before they reach
mac80211's monitor interface, which defeats the point of monitor mode
for passive capture / auditing use cases.

Follow-up to #18. Original report there leaned on kernel-level
capture/drop counters, which turned out not to move either way — see
the controlled A/B below for the metric that actually shows the effect.

## Testing

Same MT7925 hardware, same AP environment, everything else held still.
Two runs per side, `hcxdumptool -i wlp7s0 --rcascan=active --rds=5 -F`
(full 140-channel sweep across 2.4/5/6GHz), each run to completion.

| Run | drop_err | Packets captured | Dropped by kernel | APs discovered |
|---|---|---|---|---|
| Stock 1 | 1 | 50574 | 0 | 59 |
| Stock 2 | 1 | 36099 | 0 | 60 |
| Patched 1 | 0 | 41512 | 0 | 319 |
| Patched 2 | 0 | 41162 | 0 | 348 |

Kernel capture/drop counters: no difference either way (0 both sides,
every run). Raw packet count: stock captured more, not fewer — not the
effect I originally expected.

AP discovery: stock consistently ~59-60, patched consistently
~319-348, repeatable across two runs each. ~5.5x increase, and it's
the metric that's actually sensitive to this flag — `drop_err=0` lets
through weak/marginal beacons and probe responses that stock firmware
filters out, and they get parsed as distinct APs.

CC @Lucid-Duck

### DIFF
--- a/mt7925/main.c
+++ b/mt7925/main.c
@@ -923,11 +923,11 @@ static void mt7925_configure_filter(struct ieee80211_hw *hw,
 				    unsigned int *total_flags,
 				    u64 multicast)
 {
-#define MT7925_FILTER_FCSFAIL    BIT(2)
 #define MT7925_FILTER_CONTROL    BIT(5)
 #define MT7925_FILTER_OTHER_BSS  BIT(6)
 #define MT7925_FILTER_ENABLE     BIT(31)
 	struct mt792x_dev *dev = mt792x_hw_dev(hw);
+	struct mt792x_phy *phy = mt792x_hw_phy(hw);
 	u32 flags = MT7925_FILTER_ENABLE;
 
 #define MT7925_FILTER(_fif, _type) do {			\
@@ -938,6 +938,12 @@ static void mt7925_configure_filter(struct ieee80211_hw *hw,
 	MT7925_FILTER(FIF_FCSFAIL, FCSFAIL);
 	MT7925_FILTER(FIF_CONTROL, CONTROL);
 	MT7925_FILTER(FIF_OTHER_BSS, OTHER_BSS);
+
+	/* Persist what mac80211 actually asked for so the firmware sniffer
+	 * config (mt7925_mcu_config_sniffer()) can honour it instead of
+	 * hardcoding drop_err.
+	 */
+	phy->rxfilter = flags;
 
 	mt792x_mutex_acquire(dev);
 	mt7925_mcu_set_rxfilter(dev, flags, 0, 0);
--- a/mt7925/main.c
+++ b/mt7925/main.c
@@ -923,7 +923,6 @@ static void mt7925_configure_filter(struct ieee80211_hw *hw,
 				    unsigned int *total_flags,
 				    u64 multicast)
 {
-#define MT7925_FILTER_FCSFAIL    BIT(2)
 #define MT7925_FILTER_CONTROL    BIT(5)
 #define MT7925_FILTER_OTHER_BSS  BIT(6)
 #define MT7925_FILTER_ENABLE     BIT(31)
@@ -938,6 +937,12 @@ static void mt7925_configure_filter(struct ieee80211_hw *hw,
 	MT7925_FILTER(FIF_FCSFAIL, FCSFAIL);
 	MT7925_FILTER(FIF_CONTROL, CONTROL);
 	MT7925_FILTER(FIF_OTHER_BSS, OTHER_BSS);
+
+	/* Persist what mac80211 actually asked for so the firmware sniffer
+	 * config (mt7925_mcu_config_sniffer()) can honour it instead of
+	 * hardcoding drop_err.
+	 */
+	dev->mt76.rxfilter = flags;
 
 	mt792x_mutex_acquire(dev);
 	mt7925_mcu_set_rxfilter(dev, flags, 0, 0);
--- a/mt7925/mcu.c
+++ b/mt7925/mcu.c
@@ -2333,7 +2333,13 @@ int mt7925_mcu_config_sniffer(struct mt792x_vif *vif,
 			.len = cpu_to_le16(sizeof(req.tlv)),
 			.control_ch = chandef->chan->hw_value,
 			.center_ch = ieee80211_frequency_to_channel(freq1),
-			.drop_err = 0,
+			/* Only pass unfiltered/marginal frames when the
+			 * monitor userspace actually asked for them
+			 * (FIF_FCSFAIL, persisted by mt7925_configure_filter()).
+			 * Otherwise keep firmware's default drop behaviour.
+			 */
+			.drop_err = !(vif->phy->rxfilter &
+				      MT7925_FILTER_FCSFAIL),
 		},
 	};
 
--- a/mt7925/mcu.c
+++ b/mt7925/mcu.c
@@ -2333,7 +2333,13 @@ int mt7925_mcu_config_sniffer(struct mt792x_vif *vif,
 			.len = cpu_to_le16(sizeof(req.tlv)),
 			.control_ch = chandef->chan->hw_value,
 			.center_ch = ieee80211_frequency_to_channel(freq1),
-			.drop_err = 0,
+			/* Only pass unfiltered/marginal frames when the
+			 * monitor userspace actually asked for them
+			 * (FIF_FCSFAIL, persisted by mt7925_configure_filter()).
+			 * Otherwise keep firmware's default drop behaviour.
+			 */
+			.drop_err = !(vif->phy->dev->mt76.rxfilter &
+				      MT7925_FILTER_FCSFAIL),
 		},
 	};
 
--- a/mt7925/mcu.c
+++ b/mt7925/mcu.c
@@ -2333,7 +2333,7 @@ int mt7925_mcu_config_sniffer(struct mt792x_vif *vif,
 			.len = cpu_to_le16(sizeof(req.tlv)),
 			.control_ch = chandef->chan->hw_value,
 			.center_ch = ieee80211_frequency_to_channel(freq1),
-			.drop_err = 1,
+			.drop_err = 0,
 		},
 	};
 
--- a/mt7925/mt7925.h
+++ b/mt7925/mt7925.h
@@ -9,6 +9,12 @@
 
 #define MT7925_BEACON_RATES_TBL		25
 
+/* Bit mac80211's FIF_FCSFAIL request is mirrored into, persisted on
+ * dev->mt76.rxfilter by mt7925_configure_filter() so mt7925_mcu_config_sniffer()
+ * can honour it later instead of hardcoding the firmware sniffer's drop_err.
+ */
+#define MT7925_FILTER_FCSFAIL		BIT(2)
+
 #define MT7925_TX_RING_SIZE		2048
 #define MT7925_TX_MCU_RING_SIZE		256
 #define MT7925_TX_FWDL_RING_SIZE	128
--- a/mt792x.h
+++ b/mt792x.h
@@ -175,6 +175,8 @@ struct mt792x_phy {
 	s16 coverage_class;
 	u8 slottime;
 
+	u32 rxfilter;
+
 	u32 rx_ampdu_ts;
 	u32 ampdu_ref;
 


morrownr:
@Lucid-Duck 

Should we check to see if this is also applicable to mt7921, nt7612 and mt7610?

I just have this feeling that this might not make it by Felix. No specific reason. Have we considered all consequences? I want this to happen so my thoughts are supportive, not negative.

morrownr:
@Ashcal9669 

Do you have your mt7921 based usb adapter yet?

Lucid-Duck:
@morrownr @Ashcal9669

Had a look. Short answers first.

mt7921, yes. Same thing, and the two functions are near enough word for word identical. Whatever we settle on here drops straight in there.

mt7612 and mt7610, no. Different era, nothing like it in them at all. They filter in a register instead.

On whether Felix takes it, I think your gut is right. The driver already asks mac80211 whether the user wants broken frames, and then the sniffer setup overrides that from firmware. The override is the real bug. But swapping a hardcoded 1 for a hardcoded 0 just picks the other side for everybody, and takes the choice away from anyone who did want them filtered.

Doing it the way the rest of the family does means following what was actually asked for. One snag: these two drivers are the only ones that work the answer out and then throw it away, so the sniffer code has nothing to read. Five of their siblings hang onto theirs. That is the bit that is missing, and it is not much.

morrownr:
Okay, so we need this project to handle mt7925 and mt7921.

I think paragraph two in the proposed patch should be replaced with one that points out specifically why this matters to security/pen analysts.

Then the change should only apply to interfaces that are in monitor mode. 

IF wireless_interface=monitor_mode
THEN  .drop_err = 0
ELSE .drop_err = 1

And there is the issue of the adapter/device having more than one interface going at the same time.

Lucid-Duck:
@Ashcal9669 @morrownr

Ash, you were right about that field. Forced it both ways inside one driver load on the MT7927, three runs each way, alternating, with a second radio listening on the same channel.

At the shipped setting nothing errored came through. Zero, all three runs. Flipped, it was 2053, 2214 and 2369. The other radio heard between 839 and 1214 every run including the zeros, so the frames were in the air.

Worth dropping the AP count though. On a noisy channel I got 286 in a minute and five of them existed. The rest were corrupted beacons each inventing an address, and 268 of the 281 were seen exactly once. That number climbs whenever broken frames get through, so it rises whether a change helped or not. Counting only the ones seen on a clean beacon fixes it.

Nick, your three. The older part does not care about this field, six runs, 1393/1217/968 against 1207/1051/814, so the spread inside a setting is wider than the gap between them. Monitor-only is nearly already how it behaves here, the two direct paths check the interface type, and the third walks every interface but keys off a hardware flag that never comes on for this chip. Several interfaces is fine either way, mac80211 keeps one filter for the whole radio rather than one each, so they cannot disagree, and with a station up beside the monitor nothing was sent for it.

Following the flag beats pinning it to zero. It leaves people the choice, where hardcoding hands broken frames to everyone including those who never asked. It is also the version Felix is likelier to take, since the driver already works out what was asked for and then overrides it.

```diff
diff --git a/mt7925/main.c b/mt7925/main.c
index 26c4734..5638d69 100644
--- a/mt7925/main.c
+++ b/mt7925/main.c
@@ -923,11 +923,8 @@ static void mt7925_configure_filter(struct ieee80211_hw *hw,
 				    unsigned int *total_flags,
 				    u64 multicast)
 {
-#define MT7925_FILTER_FCSFAIL    BIT(2)
-#define MT7925_FILTER_CONTROL    BIT(5)
-#define MT7925_FILTER_OTHER_BSS  BIT(6)
-#define MT7925_FILTER_ENABLE     BIT(31)
 	struct mt792x_dev *dev = mt792x_hw_dev(hw);
+	struct mt792x_phy *phy = mt792x_hw_phy(hw);
 	u32 flags = MT7925_FILTER_ENABLE;
 
 #define MT7925_FILTER(_fif, _type) do {			\
@@ -940,6 +937,7 @@ static void mt7925_configure_filter(struct ieee80211_hw *hw,
 	MT7925_FILTER(FIF_OTHER_BSS, OTHER_BSS);
 
 	mt792x_mutex_acquire(dev);
+	phy->rxfilter = flags;
 	mt7925_mcu_set_rxfilter(dev, flags, 0, 0);
 	if (is_mt7927(&dev->mt76) &&
 	    (hw->conf.flags & IEEE80211_CONF_MONITOR))
diff --git a/mt7925/mcu.c b/mt7925/mcu.c
index 91fdf63..3f94621 100644
--- a/mt7925/mcu.c
+++ b/mt7925/mcu.c
@@ -2333,7 +2333,8 @@ int mt7925_mcu_config_sniffer(struct mt792x_vif *vif,
 			.len = cpu_to_le16(sizeof(req.tlv)),
 			.control_ch = chandef->chan->hw_value,
 			.center_ch = ieee80211_frequency_to_channel(freq1),
-			.drop_err = 1,
+			.drop_err = !(vif->phy->rxfilter &
+				       MT7925_FILTER_FCSFAIL),
 		},
 	};
 
diff --git a/mt7925/mt7925.h b/mt7925/mt7925.h
index 321e732..fae4033 100644
--- a/mt7925/mt7925.h
+++ b/mt7925/mt7925.h
@@ -26,6 +26,11 @@
 #define MT7925_SKU_MAX_DELTA_IDX	MT7925_SKU_RATE_NUM
 #define MT7925_SKU_TABLE_SIZE		(MT7925_SKU_RATE_NUM + 1)
 
+#define MT7925_FILTER_FCSFAIL		BIT(2)
+#define MT7925_FILTER_CONTROL		BIT(5)
+#define MT7925_FILTER_OTHER_BSS		BIT(6)
+#define MT7925_FILTER_ENABLE		BIT(31)
+
 #define MCU_UNI_EVENT_ROC  0x27
 
 #define HIF_TRAFFIC_IDLE 0x2
diff --git a/mt792x.h b/mt792x.h
index 57a38d3..5559cce 100644
--- a/mt792x.h
+++ b/mt792x.h
@@ -198,6 +198,8 @@ struct mt792x_phy {
 	wait_queue_head_t roc_wait;
 	u8 roc_token_id;
 	bool roc_grant;
+
+	u32 rxfilter;
 };
 
 struct mt792x_irq_map {
```

The bit definitions move out of the function body so the sniffer path can read them. Builds clean here with no new warnings, checkpatch clean, and the same change against mainline applies there without fuzz.

Ashcal9669:
Thanks for the numbers, Lucid — confirms this needed fixing. Good news: this is already handled in the PR as pushed (2nd commit, `3b3dd3ac`, "honour mac80211's FIF_FCSFAIL instead of hardcoding drop_err") — same idea as your patch, `drop_err` is now derived from whether FIF_FCSFAIL was actually requested instead of hardcoded. I stored it on `dev->mt76.rxfilter` rather than a new field, to match how mt7996/mt7915/mt7615/mt7603 already persist their filter state, but functionally it's the same fix you're proposing. Mind taking a look at the current diff and confirming it addresses what you tested? If your version's cleaner I'm happy to swap to it.

Also, good catch on the AP-count methodology — corrupted-beacon inflation would've kept masking real regressions in future tests. Noting that for next time.

Lucid-Duck:
@Ashcal9669

Ran your head commit on the MT7927 both ways. With no monitor flags it sends drop_err 1 and nothing errored arrives, zero out of 13622 frames. Ask for fcsfail and it sends 0, and 2882 arrive out of 13376. The other radio heard about 19000 and 17000 on the same channel either way, so the zero is the filter and not a quiet minute.

Keep yours, just move the field. You said you were following mt7996, mt7915, mt7615 and mt7603, which is the right instinct, but none of those four keeps it where you put it. The first three use their own phy struct and mt7603 its own dev. mt76x02 is the only user of mt76.rxfilter.

One thing worth knowing before Felix sees it. mac80211 raises its FIF_FCSFAIL count when a monitor opens but only lowers it again for a virtual one, so anywhere NO_VIRTUAL_MONITOR is set the count never comes back down. That is MT7927 upstream and all of mt792x in this tree. Once one monitor has asked for errored frames everything after it gets them too, until the module is unloaded, so the choice you are handing back lasts only that long. I have a two line mac80211 fix that tests clean, and the two probably want to travel together.

Ashcal9669:
@Lucid-Duck thanks for actually running the numbers — 0/13622 with no monitor flags and 2882/13376 with fcsfail requested, cross-checked against the second radio hearing ~19k/17k on the same channel, is exactly the validation this needed.

Agreed on moving the field. I followed mt7996/mt7915/mt7615 for the *filter logic* but missed that all three keep the actual storage in their own phy struct rather than `mt76.rxfilter` — that field really is mt76x02-only, so putting it there was the wrong precedent even though the behavior was right. I'll move it to a `mt792x_phy`-specific field to match.

Good catch on the FIF_FCSFAIL refcount too — didn't know mac80211 only decrements it for a virtual monitor, so any `NO_VIRTUAL_MONITOR` driver (MT7927 upstream + all of mt792x here) leaks that count up permanently once anything asks for errored frames. If you've got the two-line mac80211 fix ready, happy to have them go in together — makes more sense as one changeset than landing the driver side first and hoping the core fix follows.

Lucid-Duck:
@Ashcal9669

Thanks for getting both of yours onto the list this morning.

Here is the mt7921 half, written to match yours so the two chips cannot drift apart. Measured on an MT7922 with a read-only probe on the command itself: stock sends the drop byte even while mac80211 is asking for errored frames, patched sends it only when nobody asked, and withdrawing the request puts it back.

It does not latch here. On the MT7927 in August the request stayed on once set, because mac80211 raises the filter count when a monitor comes up and skips lowering it for drivers that set NO_VIRTUAL_MONITOR, which only that chip gets. So this one stands alone.

It is your change on another chip, so I would rather you read it before it goes out.

```diff
diff --git a/drivers/net/wireless/mediatek/mt76/mt7921/main.c b/drivers/net/wireless/mediatek/mt76/mt7921/main.c
index 68a059504e83..95f7f56125ba 100644
--- a/drivers/net/wireless/mediatek/mt76/mt7921/main.c
+++ b/drivers/net/wireless/mediatek/mt76/mt7921/main.c
@@ -668,12 +668,8 @@ static void mt7921_configure_filter(struct ieee80211_hw *hw,
 				    unsigned int *total_flags,
 				    u64 multicast)
 {
-#define MT7921_FILTER_FCSFAIL    BIT(2)
-#define MT7921_FILTER_CONTROL    BIT(5)
-#define MT7921_FILTER_OTHER_BSS  BIT(6)
-#define MT7921_FILTER_ENABLE     BIT(31)
-
 	struct mt792x_dev *dev = mt792x_hw_dev(hw);
+	struct mt792x_phy *phy = mt792x_hw_phy(hw);
 	u32 flags = MT7921_FILTER_ENABLE;
 
 #define MT7921_FILTER(_fif, _type) do {			\
@@ -685,6 +681,8 @@ static void mt7921_configure_filter(struct ieee80211_hw *hw,
 	MT7921_FILTER(FIF_CONTROL, CONTROL);
 	MT7921_FILTER(FIF_OTHER_BSS, OTHER_BSS);
 
+	phy->rxfilter = flags;
+
 	mt792x_mutex_acquire(dev);
 	mt7921_mcu_set_rxfilter(dev, flags, 0, 0);
 	mt792x_mutex_release(dev);
diff --git a/drivers/net/wireless/mediatek/mt76/mt7921/mcu.c b/drivers/net/wireless/mediatek/mt76/mt7921/mcu.c
index a118a301564c..7abd1d98feed 100644
--- a/drivers/net/wireless/mediatek/mt76/mt7921/mcu.c
+++ b/drivers/net/wireless/mediatek/mt76/mt7921/mcu.c
@@ -1226,7 +1226,7 @@ int mt7921_mcu_config_sniffer(struct mt792x_vif *vif,
 			.len = cpu_to_le16(sizeof(req.tlv)),
 			.control_ch = chandef->chan->hw_value,
 			.center_ch = ieee80211_frequency_to_channel(freq1),
-			.drop_err = 1,
+			.drop_err = !(vif->phy->rxfilter & MT7921_FILTER_FCSFAIL),
 		},
 	};
 	if (chandef->chan->band < ARRAY_SIZE(ch_band))
diff --git a/drivers/net/wireless/mediatek/mt76/mt7921/mt7921.h b/drivers/net/wireless/mediatek/mt76/mt7921/mt7921.h
index fb8654b7c5e7..f7ed11353d28 100644
--- a/drivers/net/wireless/mediatek/mt76/mt7921/mt7921.h
+++ b/drivers/net/wireless/mediatek/mt76/mt7921/mt7921.h
@@ -7,6 +7,11 @@
 #include "../mt792x.h"
 #include "regs.h"
 
+#define MT7921_FILTER_FCSFAIL		BIT(2)
+#define MT7921_FILTER_CONTROL		BIT(5)
+#define MT7921_FILTER_OTHER_BSS		BIT(6)
+#define MT7921_FILTER_ENABLE		BIT(31)
+
 #define MT7921_MAX_AID                  20
 
 #define MT7921_TX_RING_SIZE		2048
-- 
```

This PR keeps the flag on the mt76 core; the version you sent upstream keeps it on the phy. Move it to the phy and I will merge it, with mt7921 following in the same shape. Everywhere else in the tree that core field holds a hardware register value, three drivers seed it by reading the register back, and matching upstream keeps a later pick off a difference nobody meant.

Ashcal9669:
@Lucid-Duck good catch, and thanks for building the mt7921 half to match. Moved it to the phy — `struct mt792x_phy.rxfilter` in `mt792x.h`, `phy->rxfilter` in both `mt7925_configure_filter()` and `mt7925_mcu_config_sniffer()`. Agreed on the reasoning: `dev->mt76.rxfilter` already means "raw register value" elsewhere in the tree, and this filter-request bit isn't that — good call catching the mismatch before it became a real inconsistency for later picks.

Rebuilt clean (all mt792x-family modules, since `mt792x.h` is shared), pushed to this branch (`4b257f44`, force-pushed over the old commit rather than stacking a fixup — figured you'd want to review the corrected version in place, not a diff-of-a-diff).

Ready for the mt7921 half whenever you want to send it — same shape now.

PR #97, opened 2026-08-30
wifi: mt76: mt7925: re-arm monitor sniffer after chip reset recovery

Ashcal9669:
mt7925_mac_reset_work() only restores STATION/AP vifs via mt7925_vif_connect_iter() after a chip reset. A chip reset reloads firmware from scratch (triggered here by an MCU command timeout, e.g. firmware wedging mid-capture: "Message 00020002 (seq 12) timeout"), which drops the firmware sniffer's armed state. A monitor vif's local netdev/wiphy config (type, channel) is untouched by the reset, so it keeps reporting healthy (`iw dev` still shows type monitor, correct channel) while producing zero packets - only a manual rmmod/insmod of the whole driver stack recovers it.

Re-arm any monitor vif the same way mt7925_assign_vif_chanctx() already does on a fresh channel assignment: check vif->type directly rather than hw->conf.flags & IEEE80211_CONF_MONITOR. This driver sets NO_VIRTUAL_MONITOR unconditionally, and mac80211's iface.c only ever sets that hw conf bit on the virtual-monitor path, which NO_VIRTUAL_MONITOR skips entirely - so the bit is always clear here and a flag-based check silently never fires for a real monitor vif (confirmed via trace: conf.flags=0x0 at the point of recovery).

## Testing

Verified on MT7927 hardware: triggered chip reset directly via the existing `chip_reset` debugfs node while a monitor vif was actively capturing real beacon/data traffic on 2.4GHz.

- Before: post-reset capture consistently returned 0 packets despite the interface still reporting type monitor and the correct channel; only a full rmmod/insmod of the driver stack restored capture.
- After: capture recovers automatically with no manual reload, across 3 consecutive reset cycles, 600-1000 real packets captured each time.

Signed-off-by: Mark Anthony Agarro <markanthonyagarro@gmail.com>


Ashcal9669:
Follow-up validation on the actual regression this fix addresses.

Our MLO test setup depends on a separate machine (3 USB monitor dongles, one per band) doing continuous simultaneous 2.4/5/6GHz capture via `hcxdumptool`/`dumpcap` throughout multi-hour MLO test sessions. Before this fix, an MCU command timeout during capture (firmware wedging mid-session) would silently kill the monitor vif until a full `rmmod`/`insmod` of the driver stack - which was breaking our test methodology, since we can't cleanly bounce the driver mid-capture without losing state.

Re-ran a fresh 3-band capture on the patched build to confirm the fix holds under real sustained load, not just the synthetic debugfs `chip_reset` trigger from the original test:

- Duration: continuous, ~5 minutes
- Capture interfaces: 3 (one per band, simultaneous)
- Total packets: 145,182
- Per-band: 2.4GHz 24,488 / 5GHz 34,306 / 6GHz 86,388
- No dropped frames on any interface, no manual driver reload needed at any point during the run

For context on why this matters beyond monitor mode itself: this capture pipeline is how we independently verify real over-the-air behavior (vs. host-side driver counters, which can lie) for our MT7927 MLO work upstream of this repo. A monitor vif that silently dies on an MCU timeout and needs a full reload to recover was a real blocker for that verification loop, not just a monitor-mode nicety.

I merged this without review since it was low-risk and time-sensitive on our end - flagging here in case anyone wants to look at it after the fact.

Lucid-Duck:
@Ashcal9669 Thanks for the three-band run, that is a much better test than the debugfs trigger, and 145k packets with no reload is the part that needed hardware.

You asked for eyes after the fact, so here is one, left over from the 29th. The device mutex is still taken around the walk rather than inside the callback, so the order is device mutex then list lock, and mac80211 goes the other way when it tears a monitor interface down.

It costs nothing in this tree, because this driver sets NO_VIRTUAL_MONITOR and that teardown returns before it ever reaches the list lock. Upstream still gates that flag on the 7927, so the path is live there and the order would need changing before this could go up.

The move is small: take the mutex inside mt7925_monitor_reset_iter, around the arm_sniffer call, which is what its own comment already asks for. It has one caller, so nothing else is affected.

Worth closing #69 too, now this is in. Thanks again for your hard work, this repo is much better for it :)

Ashcal9669:
@Lucid-Duck No problem, and thanks for the review.

Also, if you have any notes, patches, prior findings, or pointers on MLO/STR with MT7927/mt76, I’d really appreciate them. I’m currently working through MT7927E MLO and have now validated that the AP recognizes all three two-link combinations: 0x3, 0x5, and 0x6.

The remaining issue is further down the data path. Link-layer association completes, PTK/GTK installation and PHY rates can look healthy, but actual traffic can still fail. I’ve also corroborated this independently from the AP side: its internal STA tracker sees the DUT and reports repeated DNS request timeouts for the DUT’s MLO link MAC while the client simultaneously shows the data-path failure. So this is no longer just a client-side counter/display issue.

If you’ve encountered anything similar in mt76/mac80211 MLO—particularly around active-link transitions, per-link WCID/key state, TX/RX data-path handling, or STR/MLMR behavior—any notes or direction would be extremely helpful. I’m collecting reproducible traces and hardware validation as I narrow it down.

Ashcal9669:
Also worth noting: the firmware is hardcoded to support two simultaneous MLO links, so three-band aggregation (2.4 + 5 + 6 GHz) is not possible on this Wi-Fi card. That is actually good news for our implementation, because we only need to select and dynamically switch between the best two links based on signal strength and link quality, rather than dealing with more complex three-link or eMLSR (Enhanced Multi-Link Single Radio).

Lucid-Duck:
@Ashcal9669 One number before you build on it. The driver advertises one simultaneous link, not two. mt7925/main.c encodes zero into the maximum-simultaneous-links field, and that field is N minus one, which mt7996/init.c settles by writing its radio count minus one.

The two-entry struct you found is the remain-on-channel request rather than a count of links carrying traffic. In that same function the driver asks for MLSR whenever the field is clear, and MLSR is negotiate several, operate one.

So the switching you describe is already the mode. What is missing is the selection policy, not the mechanism.

Ashcal9669:
@Lucid-Duck Read your gist (https://gist.github.com/Lucid-Duck/c764de9da584ff33ec978fb1be858962) — two sections overlap with fixes we already have on our MT7927 MLO branch, wanted to flag before you spend more time on them.

**STR advertisement, both directions**: same root cause as your PR #53 analysis. `mt7925_mcu_sta_eht_mld_tlv()` had either/or logic — STR capability only got set in the `else` branch when `EML_CAP_EMLSR_SUPP` was clear, so any chip reporting EMLSR support (this one does) never advertised STR at all, regardless of whether it can actually do simultaneous dual-link TX (it can — proven repeatedly under test). Fix: always set `str_cap`, independent of the EMLSR bit. Verified live via a `mlo_str_cap` debugfs node we added: `STR-advertised, byte0=0x02` post-fix. Matches your reconnect-failure signature (`join path`, firmware refusing immediate reconnects post-reset) — this is very likely the same bug from the other side.

**The multi-link stall (5+6 pair)**: different mechanism than what you bisected (we haven't dug into the station-tag-builder refactor you found), but same failure class — 5+6 GHz traffic silently dying with the link still "up." Ours traced to firmware DBDC allocator: both links in a JOIN batch requested the same `0xfe` (auto-DBDC) sentinel, letting firmware grant both the *same physical dbdcband* — confirmed live via the ROCEV GRANT trace (both links returning `dbdcband=1`). Fix: request distinct explicit `dbdcband` per link instead of the shared auto value. Before: secondary link `tx_pkts` incrementing but `cum_tx_airtime_us` flat 0 all day. After: real airtime (~5.2M us), ping 0/N → 20/40. Doesn't fully resolve reliability yet (~50% loss remains on that pair), but it's a real, separate contributing fault worth ruling out before chasing the builder-refactor bisect on a given failure.

On your "Not established" ask (AP-side per-link byte counters during a live association) — don't have clean counters for that yet either. Have partial AP-side signal though: `iw station dump` on our AP shows empty per-vap station tables for MLO stations (display gap, not real vacancy — the AP's internal STA_TRACKER still logs the client, e.g. repeated DNS-timeout entries for its MAC), and per-radio hostapd.conf confirms which SSID actually has a live MLD group (`upper_mld`/`lower_*` sysfs) vs which doesn't. Might be useful groundwork if you go after the per-link counters directly.

Lucid-Duck:
@Ashcal9669 Thanks for the str_cap trace. The gist was everything I had on MLO, so pull whatever's useful out of it.

str_cap checks out. str_cap[0] is written in that else branch and nowhere else in the driver, and the TLV is zeroed on add, so once EMLSR_SUPP is set STR goes out unadvertised. Two of us landing there independently is a good sign. Worth sending up on its own.

The 0xfe I read differently. It only goes out when the type is JOIN, and mt7925_mcu_set_mlo_roc picks JOIN only when MAX_SIMUL_LINKS is non-zero, which mt7925_init_mlo_caps hardcodes to zero. You're seeing 0xfe, so something on your branch is setting that field, and a maintainer will ask.

That leads somewhere you may want. The grant handler only iterates for reqtype JOIN, so stock never reaches mt7925_mcu_roc_iter here at all. Once you enable JOIN it does, and then it returns early for MLD station vifs before storing band_idx. So the grant comes back and still nothing records which radio the firmware picked.

One to skip: the request type is band-derived, AA when neither link is 2.4 GHz, matching my stall matrix exactly. I forced it both ways today. AG on 5+6 never reaches two links; AA on 2.4+5 reaches two and drops them. Dead end.

PR #101, opened 2026-09-06
The other half of the FCS-fail fix, for mt7921

Lucid-Duck:
Accepted upstream, still waiting in Felix's mt76-fixes. Merging here puts it in front of users before a released kernel has it.

This tree already carries the mt7925 half. This is mt7921, and the field it needs is already here.

A monitor asks to see frames that failed their checksum. The driver works the answer out, sends it to the firmware, then throws it away and tells the sniffer to drop them anyway. At the edge of range that is most of what arrives.

Measured on an MT7922 here, reading the byte as the driver sends it:

```
                  drop_err            stored filter
request       stock    patched     stock       patched
none          1 1 1    1 1 1       0x00000000  0x80000040
fcsfail       1 1 1    0 0 0       0x00000000  0x80000044
none again    1 1 1    1 1 1       0x00000000  0x80000040
                                   channels 1, 6 and 11
```

Withdrawing the request puts the byte back. Nothing sticks. Stock computes the filter correctly and does not keep it.

Built unpatched and patched against each of these:

```
6.18.37-lts
7.0.5
7.1.0-rc4
7.1.10
7.2-rc4
7.2.0-rc5
```

Nine lines added, six removed. Every module, no errors, and the build log is identical to the unpatched tree. Only the four mt7921 modules change. Recorded in skip-picks section 3, tally unchanged.

### DIFF
--- a/mt7921/main.c
+++ b/mt7921/main.c
@@ -672,12 +672,8 @@ static void mt7921_configure_filter(struct ieee80211_hw *hw,
 				    unsigned int *total_flags,
 				    u64 multicast)
 {
-#define MT7921_FILTER_FCSFAIL    BIT(2)
-#define MT7921_FILTER_CONTROL    BIT(5)
-#define MT7921_FILTER_OTHER_BSS  BIT(6)
-#define MT7921_FILTER_ENABLE     BIT(31)
-
 	struct mt792x_dev *dev = mt792x_hw_dev(hw);
+	struct mt792x_phy *phy = mt792x_hw_phy(hw);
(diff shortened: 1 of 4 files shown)


Ashcal9669 (approved):
Verified diff: 9 insertions / 6 deletions in code (main.c/mcu.c/mt7921.h), matches PR description. Mirrors the mt7925 rxfilter fix already merged and live in this tree (phy->rxfilter stored in configure_filter, drop_err derived from FIF_FCSFAIL bit in config_sniffer instead of hardcoded 1). Approving.

PR #102, opened 2026-09-08
Build with clang when the kernel was built with clang

Lucid-Duck:
A kernel built with clang records that in its config, and kbuild then hands
clang-only flags to whichever compiler builds a module against it. gcc rejects
them and the build dies with `unrecognized command-line option
'-mstack-alignment=8'`, which is #100.

The Makefile now reads that config and switches to clang on its own, so
`make`, `install-driver.sh` and dkms all get it. The install script names the
three tools a clang kernel needs when they are missing. dkms 3.2 and later
already pass `LLVM=1` themselves; the non-dkms path and older dkms did not.

Measured on the reporter's kernel, CachyOS `7.2.3-1-cachyos`, clang 22.1.8:

	                                    before          after
	make                                24 errors, 0    26 modules
	install-driver.sh, no dkms          fails as #100   26 modules
	install-driver.sh, dkms 3.4.3       26 modules      26 modules
	clang missing                       fails as #100   names clang, lld, llvm

On a gcc kernel, 6.18.37, the detection stays quiet and gcc builds all 26.

Closes #100

Issue #83, opened 2026-08-15
mt7921u GTK-rekey wedge


smelnichenko:
**mt7921u STA wedges at GTK rekey under TX load — supplicant stuck in EAPOL `sendto()`; `disable_usb_sg=1` resolves it (26/26 rekeys clean)**

Adding a data point with a kernel-stack capture that narrows the mechanism, plus a mitigation result.

**Setup**

- Raspberry Pi 4 Model B, Raspberry Pi OS (Debian 13/trixie), kernel `6.18.34+rpt-rpi-v8` (arm64)
- Netgear A8000 (`0846:9060`), `mt7921u`, on the VL805 SuperSpeed port (`Bus 002 ... 5000M`)
- Adapter firmware: HW/SW `0x8a108a10`, WM firmware build `20260224110949`
- wpa_supplicant 2.10 (`2:2.10-24`), NetworkManager 1.52.1, Wi-Fi powersave **off**
- AP: hostapd, WPA3-SAE/WPA2 mixed, `wpa_group_rekey=86400`
- Workload: STA mode, sustained TCP **TX from the Pi** (an SDR IQ server streaming 8-320 Mbps)

**Failure signature**

Every daily GTK rekey that landed while the link carried TX load wedged the interface. Kernel side:

```
wlan0: deauthenticated from <ap> (Reason: 16=GROUP_KEY_HANDSHAKE_TIMEOUT)
```

wpa_supplicant never sent the group-key handshake reply, and after the deauth it is *permanently stuck*. Its kernel stack at capture time (process state S, single-threaded, unresponsive to D-Bus):

```
[<0>] sock_alloc_send_pskb+0x15c/0x2c0
[<0>] packet_sendmsg+0x6a8/0x13c8
[<0>] __sock_sendmsg+0x64/0xc0
[<0>] __sys_sendto+0x120/0x1a0
[<0>] __arm64_sys_sendto+0x30/0x48
[<0>] invoke_syscall+0x4c/0x100
[<0>] el0_svc_common.constprop.0+0x48/0xf0
[<0>] do_el0_svc+0x24/0x38
[<0>] el0_svc+0x38/0x120
[<0>] el0t_64_sync_handler+0xa0/0xe8
[<0>] el0t_64_sync+0x198/0x1a0
```

[...]

PR #103, opened 2026-09-08
wifi: mt76: fix memory leak in USB TX path

Lucid-Duck:
Eason Lai's fix for issue 83, at v4 on the list. The commits are his.

A frame too short to send is refused at the USB queue and never freed, so the sending socket stays charged. Enough of those and it blocks. That is where the reporter's supplicant was stuck. The second patch frees the frame on each refusal, as the DMA path does. Without the first, the padding helper frees it twice.

Tested on an MT7921AU, Pi 5, 6.18.34:

	                       main               this branch
	0-byte frames sent     111, then blocked  3000, no block
	1-byte control         3000               3000
	iperf3 to the AP       701/699            700/699 Mbit/s, 0 retransmits

Hans Buchheim confirmed it on a Netgear mt7921u on 7.1.9.

It does not close the report. That wedge cleared with scatter-gather off, and this exit runs first. Separate work.

Rebase merge, not squash. Squashing would put my name on Eason's commits.

### DIFF
--- a/skip-picks.txt
+++ b/skip-picks.txt
@@ -53,3 +53,10 @@ f6cf5faf  # firmware: update MT7996 firmware to version 20260311
 #   Merged 2026-09-06. Accepted upstream, waiting in Felix's mt76-fixes.
 #   On the list since 2026-08-26.
 #   https://lore.kernel.org/linux-wireless/20260826231029.39632-1-lucid_duck@justthetip.ca/
+
+# mt792x: fix UAF in SDIO TX path when out of memory
+# mt76: fix memory leak in USB TX path
+#   Eason Lai's pair, carried 2026-09-08 for issue 83. On the list since
+#   2026-08-05, at v4 since 2026-08-27.
+#   https://lore.kernel.org/linux-wireless/20260827083526.1621600-2-eason.lai@mediatek.com/
+#   https://lore.kernel.org/linux-wireless/20260827083526.1621600-3-eason.lai@mediatek.com/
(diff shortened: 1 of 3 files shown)


Lucid-Duck:
@Ashcal9669 Thanks for the #101 approval, and for checking the diff yourself.

Two small ones need a second pair of eyes when you have a moment. This one carries Eason Lai's pair from the list for the USB TX leak behind #83, commits unchanged. Rebase merge please, so the authorship stays intact.

The other is #102, the clang kernel build fix. The reporter has confirmed a clean build on their own machine since it went up.

Ashcal9669:
Reviewed the diff. Both hunks check out:

- `tx.c`: switching to `__skb_pad(last, pad, false)` moves free-on-failure ownership to the caller — matches the double-free the description calls out (padding helper freeing what the caller also frees).
- `usb.c`: the three previously-silent failure paths in `mt76u_tx_queue_skb()` now actually complete the skb instead of dropping it. Queue-full and `tx_prepare_skb()` failures route through `ieee80211_tx_status_ext()` under `spin_lock_bh(&dev->rx_lock)` — verified that's the existing locking convention for this call elsewhere in `tx.c`, not something new. `mt76u_tx_setup_buffers()` failure uses `mt76_tx_complete_skb(dev, 0xffff, ...)` instead, with a comment explaining why (already idr-registered by that point, sentinel wcid keeps the non-AQL airtime counter balanced) — that's a real, specific reason, not hand-waving.

Numbers in the description (111→3000 0-byte frames, iperf3 unaffected, independent repro on different hardware) plus the v4 mailing-list links in `skip-picks.txt` are enough for me. No concerns. Rebase-merge noted.

Issue #105, opened 2026-09-09
Failure to install mt76 driver


FoxxAngel-hub:
After getting issues with my mt7902 chipset, I tried to install the driver but this failed unexpectedly
I've been trying to figure out the issue but I'm not that well versed in this subject.
I linked my entire diagnostic file along with the output from the installation and the make log file.

[mt76-diag.txt](https://github.com/user-attachments/files/32026898/mt76-diag.txt)

Ashcal9669:
Hi, thank you for providing the complete diagnostic and build log.

I can see the problem. This is not an MT7902 detection issue; your MT7902 is being detected correctly.

The build is failing because the current mt76 source uses a newer mac80211 hardware flag:

```
IEEE80211_HW_NO_VIRTUAL_MONITOR
```

Your system is running Debian 13 with kernel 6.12.107+deb13-amd64, and that kernel's mac80211 headers do not provide this flag. The failure occurs here:

```
mt792x_core.c:760:
ieee80211_hw_set(hw, NO_VIRTUAL_MONITOR);
```

So this appears to be a kernel compatibility issue in the out-of-tree driver rather than a problem with your card or installation procedure.

The repository is intended to support the 6.12 series, so we should handle this through the compatibility layer instead of requiring you to change hardware.

One additional note: Secure Boot is enabled on your system. That is not what caused this particular failure, because compilation fails before module loading, but once the build succeeds the DKMS module will still need to be correctly signed/enrolled before it can load.

Thank you for including the complete make.log; it identified the failure immediately.

FoxxAngel-hub:
Got it working now, thanks a lot !

PR #107, opened 2026-09-13
Guard NO_VIRTUAL_MONITOR where the kernel lacks it, and correct the mt7996 threshold

Lucid-Duck:
@Ashcal9669 two guards for the same flag, one on your mt792x line and one on a wrong threshold of mine. Does this look right?

#105 first. `NO_VIRTUAL_MONITOR` is set unconditionally in mt792x, and it arrived in 6.13, so the tree does not build on anything older. The reporter is on Debian 13 with 6.12.107, and mainline does not bind MT7902 until 7.1, so this repo is their only driver.

The second is mine: mt7996 guards at 6.15, so 6.14 lost the flag here.

```
flag added to mac80211      6.13   9d40f7e32774
upstream mt7996 adopts it   6.14   69d54ce7491d
```

Built:

```
kernel                 errors  warns  undefined  modules  loaded
Debian 13, 6.12.101         0      0          0       26   26/26
Fedora, 7.2.3               0      0          0       26     n/a
```

Across every release only two cells change: mt792x on 6.12 builds, mt7996 on 6.14 gains the flag.

Not marked Closes: Secure Boot is on for the reporter, so the module still needs signing and enrolling first.

### DIFF
--- a/mt792x_core.c
+++ b/mt792x_core.c
@@ -757,7 +757,10 @@ int mt792x_init_wiphy(struct ieee80211_hw *hw)
 	 * MT7927 path. Verified on hardware: active monitor now gets a real
 	 * chanctx and receives beacons at the same rate as passive monitor.
 	 */
+/* compat: IEEE80211_HW_NO_VIRTUAL_MONITOR added to mac80211 in kernel 6.13 */
+#if LINUX_VERSION_CODE >= KERNEL_VERSION(6, 13, 0)
 	ieee80211_hw_set(hw, NO_VIRTUAL_MONITOR);
+#endif
 	ieee80211_hw_set(hw, SUPPORTS_PS);
 	ieee80211_hw_set(hw, SUPPORTS_DYNAMIC_PS);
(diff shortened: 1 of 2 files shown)


Ashcal9669:
Verified both compat claims against actual upstream history:

- `9d40f7e32774` ("wifi: mac80211: add flag to opt out of virtual monitor support", Oct 2024) is where `IEEE80211_HW_NO_VIRTUAL_MONITOR" was added to mac80211 — timing matches 6.13.
- `69d54ce7491d` ("wifi: mt76: mt7996: switch to single multi-radio wiphy", Jan 2025) is where upstream mt7996 actually replaced `WANT_MONITOR_VIF` with `NO_VIRTUAL_MONITOR" — timing matches 6.14, not 6.15.

Also confirmed on current `main`: the `mt792x_core.c` call is presently unconditional (no guard at all), matching the build failure reported in #105 on 6.12.107, and `mt7996/init.c`'s existing guard is presently set to 6.15, one release later than upstream actually adopted the flag.

Small, correctly scoped, and matches the real kernel history. LGTM.

